PR #4, opened 2017-12-12
Button Component

dcorb:

_None_

**New Features**

- `Button` component
- `IconArrow` component

**Bug Fixes** 

_None_


src/components/Button/Button.js
```diff
@@ -0,0 +1,97 @@
...
+import cx from 'classnames';
+import './Button.css';
+
+const propTypes = {
+    /** Extra classes for the top-level wrapper */
+    className: PropTypes.string,
+    /** Expand width 100% */
+    isFluid: PropTypes.bool,
+    isActive: PropTypes.bool,
+    /** will show opaque and won't react to clicks because of "pointer-events:none" */
```
jfrej: Using `pointer-events:none` is not really accessible. It might make sense to add the `disabled` attribute to the `<button>` element and let the browser handle the interactions.

jimmed: :point_up: the browser isn't reliable in its handling of `disabled` - normally `onClick` events will *still* fire.

dcorb: component already adds "disabled" to the component, and we already check "isDisabled" flag to decide if props.onClick is called or not

Based on this, I will remove pointer-events:none, in case we still want to show a tooltip onHover for example for disabled status.

src/components/Button/Button.css
```diff
@@ -1,108 +1,119 @@
 .uir-button {
-    position:relative;
-    border:0; /* override browser's default */
-    color:rgba(47,40,51,0.6);
-    min-width:50px;
+    position: relative;
+    border: 0; /* override browser's default */
+    color: rgba(47,40,51,.6);
```
jimmed: There's an immediately obvious case for using SASS (or another CSS preprocessor) here, in that we're repeating the colour many, many times (albeit with different opacity).

In SASS this could be:

```sass
$buttonGrey: rgba(47, 40, 51);

$buttonFgDefault: transparentize($buttonGrey, 0.6);
$buttonBgDefault: transparentize($buttonGrey, 0.05);

$buttonFgHover: $buttonGrey;
$buttonBgHover: transparentize($buttonGrey, 0.1);
```

jfrej:
I'm afraid you'll have to modify your tests slightly - we're now using chai assertions. So for example `.toBe()` becomes `to.be()`

You can refer to: https://devhints.io/chai

.storybook/config.js
```diff
@@ -19,7 +19,7 @@ setOptions({
    *   /\./ - split by `.`
    *   /\/|\./ - split by `/` or `.`
    */
-  hierarchySeparator: null,
+  hierarchySeparator: /\./,
```
jimmed: Awesome! Didn't know storybook did this <3

src/components/Button/Button.scss
```diff
@@ -0,0 +1,67 @@
...
+    font-family: "DINOT", sans-serif;
+    box-sizing: border-box;
+    font-size: 12px;
+    vertical-align: middle;
+    user-select: none;
+    outline: none;
+    background-color: rgba(47, 40, 51, .05);
+}
+
+.uir-Button-clear {
```
jimmed: If we're going to use SASS, we should take advantage of its nice features, like rule nesting:

```sass
.uir-Button {
	...
	&-clear {
		...
	}
	&:hover {
		...
	}
}
```

src/components/Button/Button.scss
```diff
@@ -0,0 +1,67 @@
...
+    background: transparent;
+    color: black;
+}
+
+.uir-Button:hover {
+    background-color: rgba(47, 40, 51, .1);
+    color: rgb(47,40,51);
```
jimmed: Also colourisation variables/functions:

```sass
$grey: rgb(47, 40, 51);
...
&:hover {
	background-color: transparentize($grey, .1);
	color: $grey
}
```

src/components/Button/Button.story.js
```diff
@@ -0,0 +1,66 @@
...
+);
+
+
+stories.add(
+    'with icon',
+    withInfo(`
+        Icons, or other elements passed as children could need small adjustments to centre them vertically
+    `)(() => (
+        <div>
+            <Button isClear>Clear with Icon <IconArrow style={{ marginTop: '-3px' }} /></Button>
```
jimmed: When specifying pixel values in React, you should always use integers:

```js
{ marginTop: -3 }
```

src/components/Button/Button.scss
dcorb: Agree, I converted css to sass file without using any sass powers

These are variables I created for previous m-dash-ui (top 40 lines)
https://github.com/Mudano/m-dash-ui/blob/master/css/vars.scss 
The fact that designers work with semi-transparent backgrounds (maybe in the near future maybe even blur backgrounds), makes it a bit more difficult to calculate the correct backgrounds for all the variations. I used sass `rgba` functions to apply opacity to hex colors

Color variables need to get approved by Antoine / Makis

src/components/Button/Button.story.js
dcorb: good to know

src/components/Button/Button.scss
dcorb: I prefer `rgba` for this reason. transparentize is a false-friend. These 2 are equivalent:

`
transparentize(#FFF, .9);
rgba(#FFF, .1);
`

src/components/Button/Button.js
```diff
@@ -0,0 +1,93 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import ButtonType from './ButtonType';
+import './Button.scss';
+
+const propTypes = {
+    /** Content of the button, e.g. a string label or more complex React components */
```
andyberry88: I'm not sure we need the comments here. The naming of the props should be enough to indicate how/where they are used.

src/components/Button/Button.js
```diff
@@ -0,0 +1,93 @@
...
+    className: '',
+    isActive: false,
+    isDisabled: false,
+    isFullWidth: false,
+    onClick: null,
+    style: {},
+    type: ButtonType.DEFAULT,
+};
+
+class Button extends Component {
```
andyberry88: Are we going to support the confirm/reject behaviour in the existing button?

src/components/Button/Button.js
jfrej: Removed

src/components/Button/Button.js
jfrej: Added `hasConfirm` to support the confirmation state

src/components/Button/Button.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    handleConfirmClick = (event) => {
+        event.stopPropagation();
+        this.setState({
+            confirmed: true,
+            confirming: false,
+        }, () => {
+            // Wait 1s for animation to finish
+            this.confirmedTimeout = setTimeout(() => {
```
mattpauldavies: I think we should keep this for now and then reconsider all our animation options later on...

This animate off via the timeout varies from what we've done with Tooltip

src/components/Button/Button.js
```diff
@@ -0,0 +1,204 @@
+import React, { Component } from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import ListPropType from '../../prop-types/list';
+import { ButtonIconPosition, ButtonType, ButtonVariant } from './ButtonEnums';
```
mattpauldavies: I like this a lot – I'm going to apply the same concept to the Avatar collection of components

jfrej:
Original PR description by @dcorb :

* I placed Proptypes and defaultProps as consts on top of the file to keep the same pattern across all components. If we do static proptypes it won’t be supported by stateless components (in that case they  will go to the bottom)
* `isDisable` prop name style for some booleans instead of `disabled`, to make clear they are booleans.
* All components should support `classname` and `style` via props. 
* using https://github.com/JedWatson/classnames as `cx`
* Binding in constructor, instead of arrow functions, for performance reasons (ui-react components are expected to be by dozens in a single page)

src/components/Button/Button.js
jfrej: Agreed. It would be better to avoid the timeout.

PR #9, opened 2017-12-14
Adjust codecept to ensure it runs

mattpauldavies:
Tweaked codecept config to fit with new dev tooling
Also added some documentation


README_CODECEPT.md
```diff
@@ -6,22 +6,31 @@ This document gives the user an insight of how to use CodeceptJS, including pre-
 
 ### CodeceptJS Installation and Setup
 
-NB: `yarn testint` is always an alias for the `codecept` command.
+NB: `yarn test:int` is an alias for `codeceptjs` command.
```
rizmudano: I don't think this is needed anymore. Might be worth mentioning that `yarn codeceptjs` needs to be used instead of `codeceptjs` but if not, then it can be removed entirely

mattpauldavies:
Well spotted – thanks for adjusting!

I think we're good to go if you want to mark as reviewed

PR #11, opened 2017-12-14
Improving build tooling

mattpauldavies:
Streamlined tooling, added in webpack to build the main library (incl. SCSS) and adjusted npm scripts to make them more explicit and user-friendly


.storybook/webpack.config.js
```diff
@@ -9,27 +10,32 @@ module.exports = (storybookBaseConfig, configType) => {
...
         options: {
-          importLoaders: 1,
+            importLoaders: 1,
         }
-      }
+      },
+      "sass-loader"
     ],
-    include: path.resolve(__dirname, '../')
+    include: path.resolve(__dirname, '../src')
```
dcorb: I wonder if it's worth it to add also postcss with autoprefixer,  like the default config of storybook's wepback does https://github.com/storybooks/storybook/blob/master/app/react/src/server/config/defaults/webpack.config.js
(then importLoaders should be 2)
Anyway, this can be in a separate PR

.storybook/webpack.config.js
```diff
@@ -9,27 +10,32 @@ module.exports = (storybookBaseConfig, configType) => {
...
   });
 
   storybookBaseConfig.module.rules.push({
     test: /\.js$/,
     enforce: "pre",
-    exclude: /node_modules/,
```
dcorb: 👍

package.json
```diff
@@ -67,21 +64,24 @@
     "eslint-plugin-react": "7.4.0",
     "jest": "^21.2.1",
     "jest-enzyme": "^4.0.1",
-    "node-sass-chokidar": "^0.0.3",
+    "node-sass": "^4.7.2",
```
dcorb: chokidar seems more performant, and is part of CRA recommendation to setup Sass
But we can go back to "node-sass", and change to chokidar later, I see value too on consuming the original library too, until we find a good reason. It would be an easy switch

src/components/DemoComponent/DemoComponent.test-int.js
```diff
@@ -1,4 +1,4 @@
-/// <reference path="./test/codecept/steps.d.ts" />
+// / <reference path="./test/codecept/steps.d.ts" />
```
dcorb: linter is trying to fix this, but it's not a mistake according to Rizwan?  I think is a TypeScript directive. Wait a minute until we clarify

webpack.config.js
```diff
@@ -0,0 +1,27 @@
...
+    },
+    module: {
+        rules: [{
+            test: /\.scss$/,
+            use: [{
+                loader: 'style-loader',
+            }, {
+                loader: 'css-loader',
+            }, {
+                loader: 'sass-loader',
```
dcorb: postcss to be added in a follow-up, and more tweaks
We can take some inspiration of production settings from well documented CRA webpack config:
https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.prod.js

.storybook/webpack.config.js
mattpauldavies: Good idea, I'll look to do that in a future PR  👍🏼

package.json
mattpauldavies: Yes, chokidar is part of CRA recommendation.  I went with standard node-sass as it came with the webpack-plugin... agreed, is an easy switch later 😊

PR #12, opened 2017-12-15
Replace Jest with Karma

jfrej:


karma.config.js
```diff
@@ -0,0 +1,84 @@
...
+        // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
+        browsers: ['ChromeHeadless'],
+
+        // Continuous Integration mode
+        // if true, Karma captures browsers, runs the tests and exits
+        singleRun: true,
+
+        // Concurrency level
+        // how many browser should be started simultaneous
+        concurrency: Infinity,
```
jimmed: We _might_ want to consider a lower number here...

mattpauldavies: Is that something you think should be changed now @jimmed?

If not I think this looks good...

PR #14, opened 2017-12-18
[MDU-43] 👕 LINT: add property order rules to stylint

jfrej:
Based on Twitter Bootstrap


mattpauldavies:
This looks good, excellent idea.  I'll get in merged in so we can pull it into the button and avatar comps

PR #18, opened 2017-12-21
New Avatar & Tooltip Components

mattpauldavies:

_None_

**New Features**  
 

- `Avatar` component
- `AvatarTitle` component
- `AvatarCard` component
- `Tooltip` component
- `TooltipBox` component

**Bug Fixes**  
 
_None_

**Performance Improvements**  

_None_

**Miscellaneous**  

- New StoryBook documentation


.eslintrc
```diff
@@ -12,13 +12,11 @@
     },
     "rules": {
         "indent": ["error", 4],
+        "max-len": ["error", 150],
```
jfrej: Why this change? The [StyleGuide](https://docs.google.com/document/d/1U8DqwOyYlNweZSa5Kn_sJW9Kx079vMclYwVi8yyPYaY/) doesn't mention this deviation from Airbnb.

.stylelintrc.js
```diff
@@ -218,7 +218,7 @@ module.exports = {
     ],
 
     // Sass rules
-    'max-nesting-depth': 2,
+    'max-nesting-depth': 3,
```
jfrej: This might be an OK change but the [StyleGuide](https://docs.google.com/document/d/1U8DqwOyYlNweZSa5Kn_sJW9Kx079vMclYwVi8yyPYaY/) should be amended to include this.

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+
+### Installation
+
+    yarn add @mudano/ui-react
+
+_NB: installation from NPM is not yet available_
+
+### Requirements
+
+UI React components are only compatibly with a React application and require the following:
```
jfrej: Typo, should be: "compatible"

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+
+### Usage
+
+You can consume UI React in one of two ways, as a whole library or as individual components.
+
+    import { YourComponent } from '@mudano/ui-react'; // UMD Modules
+
+    import { YourComponent } from 'node_modules/@mudao/ui-react/es/index'; // Native ES6 Modules
+
+    import YourComponent from 'node_modules/@mudao/ui-react/es/component/YourComponent/YourComponent';
```
jfrej: I'm pretty sure this should be `/components/` (with "s" at the end)

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+
+### Issues
+
+Spotted an issue?
+
+Please report it at [https://github.com/Mudano/ui-react/issues](https://github.com/Mudano/ui-react/issues)
+
+### Development
+
+To help with development, clone the repoistory from here: [https://github.com/Mudano/ui-react](https://github.com/Mudano/ui-react)
```
jfrej: Typo, should be "repository"

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+import sinon from 'sinon';
+import Avatar from './Avatar';
+
+describe('Avatar', () => {
+    const sampleData = {
+        name: 'Matt Davies',
+        src: 'https://randomuser.me/api/portraits/lego/2.jpg',
+    };
+
+    const sandbox = sinon.sandbox.create();
```
jfrej: `sandbox` is now available in unit test utils, so you can:
`import { sandbox } from '../../../test/unit/utils';`
and you won't need the `beforeEach` and `afterEach`

src/components/About.story.js
mattpauldavies: Thanks! Changed...

src/components/About.story.js
mattpauldavies: Thanks! I've changed

src/components/About.story.js
mattpauldavies: Thank you! I've fixed

.eslintrc
mattpauldavies: I changed this because a couple of tests were testing long strings and needed a longer `max-len`.

I've since found a much better way of running the tests.  I've undone this change and fix any linting issues.

Thanks for flagging.

.stylelintrc.js
mattpauldavies: Good point – I've added a note to the style guide...

We need a nesting depth of 3 to make the most of SASS features e.g. looping over a map of class variants to create classes

src/components/Avatar/Avatar.test.js
mattpauldavies: I intentionally didn't use your utility function, because I needed the `beforeEach` and `afterEach` to be defined within the `describe`.

If they were defined outside of the `describe` I ran into issues with my spies.  Particularly when I was stubbing console.error

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+
+    yarn add babel-core babel-loader css-loader node-sass sass-loader style-loader webpack --dev
+
+### Usage
+
+You can consume UI React in one of two ways, as a whole library or as individual components.
+
+    import { YourComponent } from '@mudano/ui-react'; // UMD Modules
+
+    import { YourComponent } from 'node_modules/@mudao/ui-react/es/index'; // Native ES6 Modules
```
dcorb: mudao

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+
+const stories = storiesOf('About', module);
+
+stories.add(
+    'Welcome',
+    overviewWrapper(`
+
+UI React is the Mudano component library.
+
+If you have questions ask the development team on Slack: [#ui-react](slack://channel?id=C89RXMYJY&team=T065RT3A7)
```
dcorb: only Mudano people can access this Slack channel, should we add this line?

I guess we also direct people to open issues in github (instead of Jira) for the same reason

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,110 @@
+import React, { Component } from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import InitialsPropType from '../../prop-types/initials';
+import ListPropType from '../../prop-types/list';
```
dcorb: didn't know about ListPropType, good one

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,110 @@
...
+    theme: ListPropType(['light', 'dark']),
+    status: PropTypes.string,
+    hasNotification: PropTypes.bool,
+    hasHalo: PropTypes.bool,
+    onClick: PropTypes.func,
+    tabIndex: PropTypes.number,
+    className: PropTypes.string,
+    style: PropTypes.objectOf(PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
```
dcorb: I like this (if it works fine for all cases), we probably should do the same style Proptypes for all style props, including Button

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,110 @@
...
+            ? <img src={this.props.src} alt={`${this.props.name} profile`} />
+            : this.props.initials || this.calculateInitials(this.props.name);
+        const notification = this.props.hasNotification
+            ? (
+                <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+                    <g fill="none" fillRule="evenodd" transform="translate(3 3)">
+                        <circle cx="9" cy="9" r="8" stroke="#F33061" strokeWidth="2" />
+                        <path fill="#F33061" d="M8 5h2v5H8zM8 11h2v2H8z" />
+                    </g>
+                </svg>
```
dcorb: does it have sense to have a separate component for this SVG, will be reused?
See Button branch, and IconArrow component

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,110 @@
...
+            )
+            : null;
+        const status = this.props.status || null;
+        const notificationOrStatus = notification || status
+            ? <span className="uir-avatar-user-status">{notification || status}</span>
+            : null;
+        return (
+            <div
+                ref={(div) => { this.avatarRef = div; }}
```
dcorb: If we are implementing ref prop in all components, would be nice to have a consistent way of declaring ref, and have always the same argument name, something like
`ref={(node) => { this.avatarRef = node; }}`

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,71 @@
+import React, { Component } from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import ListPropType from '../../prop-types/list';
+import './Tooltip.scss';
+
+const propTypes = {
+    children: PropTypes.element.isRequired,
+    tooltip: PropTypes.element.isRequired,
+    position: ListPropType(['top-left', 'top-right', 'bottom-right', 'bottom-left']),
```
dcorb: do we support / need horizontally centred tooltip? I think it's the behaviour of the previous tooltip

src/components/_variables.scss
```diff
@@ -0,0 +1,9 @@
+$font: "DINOT", sans-serif;
+
+$white: #fff;
+$grey: #ccc;
+$charocal: #3a333c;
```
dcorb: charcoal typo

src/components/Avatar/Avatar.scss
```diff
@@ -0,0 +1,182 @@
...
+            border-width: 2px;
+
+            &::before {
+                position: absolute;
+                top: -2px;
+                left: -2px;
+                z-index: -1;
+                display: block;
+                content: "";
+                // background-image: radial-gradient(farthest-corner at 0 0, $pink 0%, $pink 75%, $yellow 80%);
```
dcorb: remove?

src/components/_variables.scss
```diff
@@ -0,0 +1,9 @@
+$font: "DINOT", sans-serif;
+
+$white: #fff;
+$grey: #ccc;
```
dcorb: I think these colors are not the ones used by designers. Double check, we normally use bg with opacity, instead of direct hex color

src/components/_variables.scss
```diff
@@ -0,0 +1,9 @@
...
+$charocal: #3a333c;
+$purple: #9a3cbb;
+$pink: #fd2a7d;
+$teal: #46bead;
+$yellow: #fdbd2c
```
dcorb: if yellow is not being used, then remove it, until we need it ?

.storybook/manager-head.html
```diff
@@ -0,0 +1,11 @@
+<style>
+    /* hide command line tools popup button because
+        we've disabled several default actions */
+    .Pane.vertical.Pane1 button {
+        display: none;
+    }
+
+    #root > div > div {
+        background-color: #efeee8 !important;
```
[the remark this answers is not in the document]
mattpauldavies: This is styling the storybook container, I'll add a note 👍🏼

package.json
```diff
@@ -108,13 +109,11 @@
...
-    "@storybook/addon-info": "^3.2.17",
-    "@storybook/addon-knobs": "^3.2.17",
-    "@storybook/addon-links": "^3.2.17",
     "@storybook/addon-options": "^3.2.17",
+    "@storybook/addons": "^3.2.17",
     "@storybook/react": "^3.2.17",
-    "@storybook/storybook-deployer": "^2.2.0",
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "prop-types": "15.6.0",
```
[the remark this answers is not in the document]
mattpauldavies: We should have prop-types in both `peerDependencies` and `dependencies`...

- in `dependencies` as requested by storybook
- in `peerDependencies` as end-users will need prop-types to consume the native ES6 components

Are you suggesting that we change that?

src/components/About.story.js
mattpauldavies: Good spot, thanks!

src/components/Avatar/Avatar.js
mattpauldavies: ListPropType is something I created!  It's a wrapper of `oneOf` that allows us to access the list of options that we pass to it... we can then display the options in the props table :)

src/components/Avatar/Avatar.js
mattpauldavies: @jfrej found this, I believe it's been added to the Button too 👍

src/components/Avatar/Avatar.js
mattpauldavies: @jfrej and I have flagged icons as a point to discuss... I'll leave this SVG here for now as I believe the only place the notification icon will be used is in the Avatar... 

If we have a need for the SVG elsewhere then we could make it part of the Icon component

src/components/Avatar/Avatar.js
mattpauldavies: I'm not sure if every component should have a ref prop, we should just use it when we want to control something, in this instance I wanted to control the focus to make the element more accessible

happy to agree an argument name, node sounds good to me, I'll change it now

src/components/About.story.js
jfrej: Same on the line below

src/components/Avatar/Avatar.js
jfrej: It works but it doesn't really validate anything - it just checks if the `style` prop is an object and allows both string and number as object key values. So basically allows anything inside the object.
If we wanted proper `style` validation, we should use https://www.npmjs.com/package/react-style-proptype

src/components/Avatar/Avatar.scss
mattpauldavies: thanks for flagging, I need to chat with Antoine about whether we need the halo bordered and if we do, do we want yellow (as per m-dash-ui) or pink to be more on brand

.storybook/addons.js
```diff
@@ -1,4 +1 @@
-import '@storybook/addon-actions/register';
-import '@storybook/addon-links/register';
-import '@storybook/addon-knobs/register';
```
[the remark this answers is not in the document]
mattpauldavies: I can see the use case for knobs... I think this a point that requires further discussion

Should we create a story for every variation to allow us to explicitly test and demonstrate instead of relying on knobs for example

src/components/About.story.js
mattpauldavies: I'll remove Slack and direct towards GitHub issues :)

src/components/Tooltip/Tooltip.js
mattpauldavies: Good point, I'll check with Antoine 👍

src/components/_variables.scss
mattpauldavies: Thank you, fixed

src/components/_variables.scss
mattpauldavies: maybe used by halo, will check with Antoine, if yellow not required I will remove :)

src/components/_variables.scss
mattpauldavies: I'll check, I explicitly used hex over rgba because it allowed me to make use of the same halo technique used in m-dash-ui.  If we don't need the halo, I'll swap back to rgba 👍

src/components/Avatar/AvatarCard.scss
```diff
@@ -0,0 +1,47 @@
...
+    box-shadow: 0 2px 4px 0 rgba(0, 0, 0, .5);
+    transition-timing-function: ease-in-out;
+    transition-duration: 1s;
+    transition-property: all;
+
+    .uir-avatar-card-profile {
+        padding-right: 12px;
+    }
+
+    &.uir-avatar-card--has-role {
```
jfrej: This could be `&--has-role` to reduce specificity a little bit.

src/components/Avatar/AvatarCard.scss
```diff
@@ -0,0 +1,47 @@
...
+        .uir-avatar-card-name {
+            font-size: 26px;
+        }
+
+        .uir-avatar-card-job-role {
+            font-size: 16px;
+        }
+    }
+
+    &.uir-avatar-card--has-team {
```
jfrej: And this could be `&--has-team`

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,110 @@
...
+    size: 'md',
+    theme: 'light',
+    status: null,
+    hasNotification: false,
+    hasHalo: false,
+    onClick: null,
+    tabIndex: 0,
+    className: null,
+    style: null,
+};
```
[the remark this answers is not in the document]
mattpauldavies: excellent idea, I'll do that now

src/components/Avatar/Avatar.scss
```diff
@@ -0,0 +1,182 @@
...
+        cursor: pointer;
+        opacity: .8;
+    }
+
+    &:focus {
+        outline: none;
+        opacity: .8;
+    }
+
+    &.uir-avatar--disabled {
```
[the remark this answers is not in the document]
mattpauldavies: I like this, feels a bit more elegant than using `--disabled` I've tweaked the code accordingly

.storybook/addons.js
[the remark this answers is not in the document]
dcorb: Maybe it can be discussed separately, and leave the import line. I'm ok removing addon-actions, and addon-links can create links between stories... we can add it back if needed

package.json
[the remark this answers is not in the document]
dcorb: true, forget what I said

src/components/Avatar/Avatar.js
dcorb: ok

src/components/Avatar/Avatar.scss
```diff
@@ -0,0 +1,182 @@
...
+
+        &:hover,
+        &:focus {
+            cursor: default;
+            outline: none;
+            opacity: 1;
+        }
+    }
+
+    &.uir-avatar--halo {
```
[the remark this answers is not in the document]
mattpauldavies: Good point. Andy B liked this too, so I've rolled it out across each of the avatar components

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,104 @@
...
+});
+
+Scenario('Clickable', (I) => {
+    searchWithinIframe(I, KIND, 'Clickable', () => {
+        I.see('Clickable');
+        I.seeElement('.uir-avatar');
+        I.click('.uir-avatar');
+        I.seeInPopup('it works!');
+        I.acceptPopup();
+        // @TODO: test for key press, space and enter
```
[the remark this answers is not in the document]
mattpauldavies: Yes we should, I'd like your help (and @rizmudano) to get this working because the key press wasn't working

src/components/Avatar/Avatar.story.js
```diff
@@ -0,0 +1,319 @@
...
+    'Theme',
+    storyWrapper(
+        'Use the `theme` prop to set the status and notification icon background.',
+        <Avatar
+            name="David Smith"
+            initials="DS"
+            theme="dark"
+            size="lg"
+            src="https://randomuser.me/api/portraits/men/3.jpg"
+            status="😡"
```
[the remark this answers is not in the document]
mattpauldavies: Hmmm odd – I'll take a look into this!

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('complains if name is not provided', () => {
+        shallow(<Avatar />);
+        // eslint-disable-next-line no-console
+        expect(console.error.calledWithMatch('The prop `name` is marked as required in `Avatar`')).to.equal(true);
```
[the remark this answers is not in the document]
mattpauldavies: nice... thank you, will change

src/components/Tooltip/Tooltip.js
mattpauldavies: we do need a centred tooltip, I'll add this...

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+
+    it('complains if initials longer than 2 characters are used', () => {
+        shallow(<Avatar name={sampleData.name} initials="MPD" />);
+        // eslint-disable-next-line no-console
+        expect(console.error.calledWithMatch('Initials supplied to `Avatar` should be 1 or 2 characters')).to.equal(true);
+    });
+
+    it('renders a div element', () => {
+        const avatar = shallow(<Avatar name={sampleData.name} />);
+        expect(avatar.find('div').length).to.equal(1);
```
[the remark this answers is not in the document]
mattpauldavies: 👍 swapped over, ran this out to other tests too (e.g. tooltip)

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+    });
+
+    it('has correct class', () => {
+        const avatar = shallow(<Avatar name={sampleData.name} />);
+        expect(avatar.find('div').at(0).hasClass('uir-avatar')).to.equal(true);
```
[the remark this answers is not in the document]
mattpauldavies: Thanks for this, very helpful

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+            expect(hasClass).to.equal(true);
+        });
+    }
+
+    testComponentSize('small', 'xs');
+    testComponentSize('small', 'sm');
+    testComponentSize('medium', 'md');
+    testComponentSize('large', 'lg');
+
+    it('can display a halo around the avatar', () => {
```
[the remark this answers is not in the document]
mattpauldavies: I have to say I'm not 100% sure about the helper utils, they are VERY tidy and do remove duplication I prefer to see the whole test without the indirection so I'm really clear on what it does... what do you think?

src/components/Avatar/Avatar.test.js
```diff
@@ -0,0 +1,216 @@
...
+
+    it('can pass through class', () => {
+        const exampleClassName = 'example-class';
+        const avatar = shallow(<Avatar name={sampleData.name} className={exampleClassName} />);
+        expect(avatar.find('div').at(0).hasClass(exampleClassName)).to.equal(true);
+    });
+
+    it('calculates initials from name', () => {
+        const avatar = shallow(<Avatar name={sampleData.name} />);
+        expect(avatar.find('div').at(0).text()).to.equal('MD');
```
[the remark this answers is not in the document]
mattpauldavies: I've rolled out this style of assertion to every test :)

src/components/Avatar/Avatar.story.js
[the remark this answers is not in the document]
mattpauldavies: this should be resolved – please check again when you get a moment :)

src/components/Tooltip/TooltipBox.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+    beforeEach(() => {
+        sandbox.stub(console, 'error');
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('complains if chidlren not provided', () => {
```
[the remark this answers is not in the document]
mattpauldavies: fixed!

src/components/Tooltip/Tooltip.story.js
```diff
@@ -0,0 +1,112 @@
...
+            {exampleAvatar}
+        </Tooltip>,
+    ),
+);
+
+stories.add(
+    'Show',
+    storyWrapper(
+        'Pass the `showTooltip` prop to force show a tooltip.',
+        <Tooltip showTooltip tooltip={<TooltipBox>my tooltip</TooltipBox>}>
```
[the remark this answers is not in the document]
mattpauldavies: Ahhh well spotted, this was an issue with prop types being too strict, I've fixed

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,104 @@
...
+Scenario('Overview', (I) => {
+    searchWithinIframe(I, KIND, 'Overview', () => {
+        I.see('Avatar: Avatar');
+        I.see('Overview');
+        I.see('Example');
+        I.seeElement('.uir-avatar');
+        I.see('DS');
+        I.see('Source Code');
+        I.seeElement('code');
+        I.see('Props');
```
rizmudano: I wonder if we're running too many assertions in our integration tests?

Some of the steps looks like they're only testing Storybook (like `I.see('Props');`, and I'm not sure if we need to test each property, if we've already got a unit test for it.

The main concern is that the Avatar integration suite takes 50s to run - so 10 similar sized components would take a simple component library 5-10 minutes to run a full suite

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,104 @@
...
+        I.seeElement('table');
+        I.see('property');
+        I.see('value');
+        I.see('required');
+        I.see('default');
+        I.seeNumberOfElements('td', 12 * 4);
```
rizmudano: Again, AFAIK the props table is something that Storybook does out of the box, so I'm not sure if we need to test it?

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,104 @@
...
+        I.see('Initials');
+        I.see('Variations');
+        I.seeNumberOfElements('.uir-avatar', 3);
+    });
+});
+
+Scenario('Image', (I) => {
+    searchWithinIframe(I, KIND, 'Image', () => {
+        I.see('Image');
+        I.seeNumberOfElements('.uir-avatar', 5);
```
rizmudano: Since we use the same class name for avatar-initials and avatar-images, the 'Initials' and 'Image' test scenarios seem to validate the same state (beside the avatar sizes, but those class names are again generic for avatar-initials and avatar-images).

Again, a discussion to be had over our scope of unit and integration tests

src/components/Avatar/AvatarCard.test-int.js
```diff
@@ -0,0 +1,135 @@
+/* global Feature, Scenario  */
+const searchWithinIframe = require('../../../test/codecept/utils/searchWithinIframe.js');
```
rizmudano: Again, consider removing if we decide to include this in the `/test/codecept/steps_file.js`

src/components/Avatar/AvatarTitle.test-int.js
```diff
@@ -0,0 +1,99 @@
+/* global Feature, Scenario  */
+const searchWithinIframe = require('../../../test/codecept/utils/searchWithinIframe.js');
```
rizmudano: can remove after adding to `steps_file.js`

src/components/Tooltip/Tooltip.test-int.js
```diff
@@ -0,0 +1,45 @@
+/* global Feature, Scenario  */
+const searchWithinIframe = require('../../../test/codecept/utils/searchWithinIframe.js');
```
rizmudano: can remove if we add this to `steps_file.js`

src/components/Tooltip/TooltipBox.test-int.js
```diff
@@ -0,0 +1,36 @@
+/* global Feature, Scenario  */
+const searchWithinIframe = require('../../../test/codecept/utils/searchWithinIframe.js');
```
rizmudano: can remove if we add to `steps_file.js`

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,109 @@
+/* global Feature, Scenario */
+const searchWithinIframe = require('../../../test/codecept/utils/searchWithinIframe.js');
```
rizmudano: can remove if we decide to add this to `/test/codecept/steps_file.js`

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,104 @@
+/* global Feature, Scenario */
```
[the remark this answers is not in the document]
rizmudano: :+1:

src/components/Avatar/Avatar.test-int.js
rizmudano: Separately, I remember us saying that we can't verify avatar rendering at a unit test level, but we do have a test for displaying user profile image. Was that a mistaken assumption initially?

test/codecept/utils/searchWithinIframe.js
```diff
@@ -0,0 +1,11 @@
+/* global within  */
+
+// NB: we're using a NodeJS module because Codecept
+//      does not support native ES modules
+
+function searchWithinIframe(I, kind, story, searchWithin) {
```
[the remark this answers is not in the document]
mattpauldavies: I think this is a great idea, I'll implement it now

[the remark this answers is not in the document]
mattpauldavies: I've removed `mapper.js` for now, I'll make sure Button works after we merge this PR in :)

src/components/Avatar/Avatar.test-int.js
mattpauldavies: We've built a custom props table plugin (which is unit tested)... I'll bow to your expertise over whether it should be included in the scope of the int tests

My thought process was it's important for users to be able to make use of the documentation so we should check that it is being displayed...

.storybook/addons.js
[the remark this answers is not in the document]
mattpauldavies: Chatted this through with Andy and addon-knobs is going back in 👍

src/components/Avatar/Avatar.test-int.js
mattpauldavies: All integration tests have been refactored to make them simpler, reduce duplication and run faster as a result

rizmudano: I agree that we should test the rendering of the docs too. I think what you've done is a good balance :100:

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,41 @@
...
+
+const KIND = 'Avatar.Avatar';
+
+Feature('Avatar');
+
+Scenario('Storybook Documentation', (I) => {
```
rizmudano: Makes sense to split up documentation from the actual components :+1:

src/components/Avatar/Avatar.test-int.js
```diff
@@ -0,0 +1,41 @@
...
+        I.see('value');
+        I.see('required');
+        I.see('default');
+        I.seeNumberOfElements('td', 12 * 4);
+    });
+});
+
+Scenario('Avatar', (I) => {
+    I.searchWithinIframe(KIND, 'Overview', () => {
+        I.seeElement('.uir-avatar', 4);
```
rizmudano: What does the `4` specify? I thought it was a count, but when I changed the value to `3` or `5`, the test still passed.

src/components/Avatar/AvatarTitle.test-int.js
```diff
@@ -0,0 +1,33 @@
...
+        I.seeNumberOfElements('td', 6 * 4);
+    });
+});
+
+Scenario('AvatarTitle', (I) => {
+    I.searchWithinIframe(KIND, 'Overview', () => {
+        I.seeElement('.uir-avatar-title');
+        I.seeElement('.uir-avatar');
+        I.seeElement('.uir-avatar-title-name');
+        I.see('Jane Smith');
```
rizmudano: Might be better to condense lines 28 and 29 into:
```js
I.see('Jane Smith', '.uir-avatar-title-name');
```
Then you've verified more, by checking that Jane Smith is within the right element, and reduced execution time (albeit by a fragment of a second)

src/components/Avatar/AvatarTitle.test-int.js
```diff
@@ -0,0 +1,33 @@
...
+        I.seeElement('.uir-avatar-title-job-role');
+        I.see('Delivery Manager');
```
rizmudano: As above, can condense to:
```js
I.see('Delivery Manager', '.uir-avatar-title-job-role');
```

src/components/Avatar/AvatarCard.test-int.js
```diff
@@ -0,0 +1,35 @@
...
+        I.seeNumberOfElements('td', 6 * 4);
+    });
+});
+
+Scenario('AvatarCard', (I) => {
+    I.searchWithinIframe(KIND, 'Overview', () => {
+        I.seeElement('.uir-avatar-card');
+        I.seeElement('.uir-avatar');
+        I.seeElement('.uir-avatar-card-name');
+        I.see('David Smith');
```
rizmudano: Might be better to condense lines 28 and 29 into:

I.see('David Smith', '.uir-avatar-card-name');
Then you've verified more, by checking that David Smith is within the right element, and reduced execution time (albeit by a fragment of a second)

src/components/Avatar/AvatarCard.test-int.js
```diff
@@ -0,0 +1,35 @@
...
+        I.seeElement('.uir-avatar-card-job-role');
+        I.see('Sales Manager');
```
rizmudano: As above, can condense to:

```js
I.see('Sales Manager', '.uir-avatar-card-job-role');
```

src/components/Tooltip/Tooltip.test-int.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+});
+
+Scenario('Tooltip', (I) => {
+    I.searchWithinIframe(KIND, 'Wrapper', () => {
+        I.seeElement('.uir-tooltip');
+        I.moveCursorTo('.uir-tooltip');
+        I.seeElement('.uir-avatar-card');
+        I.moveCursorTo('h1');
+        I.dontSeeElement('.uir-avatar-card');
```
rizmudano: Very clever!

src/components/Tooltip/TooltipBox.test-int.js
```diff
@@ -0,0 +1,35 @@
...
+        I.seeElement('table');
+        I.see('property');
+        I.see('value');
+        I.see('required');
+        I.see('default');
+        I.seeNumberOfElements('td', 4 * 4);
+    });
+});
+
+// Scenario('Success', (I) => {
```
rizmudano: Commented out?

test/codecept/steps_file.js
```diff
@@ -1,12 +1,16 @@
...
-  return actor({
+    return actor({
 
-    // Define custom steps here, use 'this' to access default methods of I.
-    // It is recommended to place a general 'login' function here.
+        // Define custom steps here, use 'this' to access default methods of I.
+        // It is recommended to place a general 'login' function here.
 
-  });
+        searchWithinIframe: function(kind, story, searchWithin) {
```
rizmudano: :+1:

src/components/Avatar/Avatar.test-int.js
mattpauldavies: Hmmm it should be a count!  `I.seeElement` needs to be swapped to `I.seeNumberOfElements` ... will change

src/components/Avatar/AvatarCard.test-int.js
mattpauldavies: Nice!

src/components/Avatar/Avatar.story.js
```diff
@@ -0,0 +1,357 @@
...
...srV9U0xHez3ETx/rNC3n/L6+FL9u33Yr0Vw8+XaCXbFRE14z1xuogO88q8zyrnNeZ5Vxwfa8B69Lzltkt185pAPoM1bjgSGG0Z7i9d5AM4iUBQfj1rUWiDqQfKl3VnWC3DuQqbzuJ6YANehjxQfSEnIyTibS4dD0vtpLovJNMI4UxgkdWJ+lK12oJ3D8QDA/wDPhQ/FevTa5rzXG49hG22BfAID/WpYZRNYBif2RKt7v/lYsvn38eGmF1sqtTi/1II6MgNQQ2by3MQVSQyk8h5VcXkG6NXxkgn4irThBIv0pD2iB9jMm0+Icbf61F8orFK7Gjh22xbpy6CnSzjUbe7XEnDb6LebYwWtJOcL9eX7p9RR0UJAHKsUotOma4tNWiwgbC8qnQEnPlUVvGSPpRNwDDbEhSznkoHiT0oKLYkmFcJt9obVMgNG1wVweYPdHhSzxL7OZ47hrnQo1aFubW5bBU/w58PSnnh7TF0qwWIhe1Y75WHix61cNjFehFeYpGLI05No/OF9Z3VhOYLy3kglH4ZFwT/ehc1+idT02z1K37G8to508nXOPd5Ul3fsx0uV2Nvc3NvuPdXIZR8+f1p/N8EsyvNfE066j7M9Uto3ksriK8CDOzBRz7h0PzpJkR4naORGR1OGVhgg+RoNUE37cAcnwrM/atqLWPDcgjYiSV+zXHgD1+madNX1m1098tJuk69mvMmsQ9oPEcutTwwm3MUAkyNzZLePStjl5i2DzbEKaDY0mB3QQPjRtm+x54yMLJHux9M186Y7ONucg77e81FCQbp2/eQj/wAaxFiwt5AbBGbrEQvv8KK0eddO162lxmJmVh64OaAsucawN1ZDu9//AA1yrEIB+JGyuPDxx+fyoUNF0fpTTbqeVGW4RZo5SG2sOR8seVHz6XF2DzQK6bRko3MAehqs9ntwmscIWk7d54xsz7un/PSm1o8wyA5yUI+lPKEZRt9F9OMqQv28axr61aWdisuy5lXO05jB/Og7O3a8uAg/ZJgyHzHlTGHjhXDuqj1PhUccK2Nkn9I9zgc/KvVcHxHzqCWaNrd2Qhu6cYoeG0QBMHqBV6sgHyAFagI3KQeoqQpsjUZrkc1LeOKKOPk/U3HxpI464Pi1S1m1WyVlv40yyL0mA/8AbHjTtuxEPWuQcjFGrOP/2Q=='; // eslint-disable-line
```
jimmed: We can definitely use something like `image-loader` for webpack to store this file externally.

src/components/Avatar/AvatarCard.js
```diff
@@ -0,0 +1,61 @@
...
+    const profile = props.avatar
+        ? <div className="uir-avatar-card-profile">{props.avatar}</div>
+        : null;
+    return (
+        <div
+            className={cx(
+                'uir-avatar-card',
+                { 'uir-avatar-card--has-role': jobRole },
+                { 'uir-avatar-card--has-team': team },
+                { 'uir-avatar-card--has-profile': profile },
```
jimmed: These can all exist within a single object - there's no need to create 3 separate ones

src/components/Avatar/Avatar.test.js
jimmed: Arguably, we want to take advantage of `sandbox` as much as possible, but based on this, its API isn't ideal for most circumstances.

src/components/Tooltip/TooltipBox.js
```diff
@@ -0,0 +1,44 @@
...
+    children: PropTypes.oneOfType([
+        PropTypes.element,
+        PropTypes.string,
+    ]).isRequired,
+    status: ListPropType(['default', 'success', 'error']),
+    className: PropTypes.string,
+    style: PropTypes.objectOf(PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+    ])),
```
jfrej: For consistency, could you please order props alphabetically here as well?

src/storybook-addons/storyWrapper.test.js
```diff
@@ -0,0 +1,61 @@
...
+        </div>,
+    );
+    const storyContextMock = { kind: 'Foo', story: 'Bar' };
+    return mount(hoc(storyContextMock));
+}
+
+describe('storyWrapper', () => {
+    it('renders our logo', () => {
+        const wrapper = prepareWrapper();
+        expect(wrapper.find('svg').length).to.equal(1);
```
jfrej: This doesn't really test if it renders *our* logo, just a logo. Perhaps that's ok, if so the test title could be changed to "renders a logo".
Or if you want to be more thorough, you could add a proper `<title>` to the svg (currently it has some dummy text) and test for that. And you can also check the `fill` attribute on the path to see that it uses the correct brand colour (#9a3cbb).
But I'm not sure if we need to be this thorough.

src/components/Avatar/Avatar.story.js
```diff
@@ -0,0 +1,357 @@
...
+            <Avatar
+                name="David Smith"
+                src="https://randomuser.me/api/portraits/men/3.jpg"
+                style={{ marginRight: '30px' }}
+            />
+            <Avatar
+                name="Jane Smith"
+                src="https://randomuser.me/api/portraits/women/58.jpg"
+                hasHalo
+                hasNotification
```
rizmudano: @mattpauldavies If this is enough to enable the prop, then would we need to change the props table? Currently the props table shows:

property | value | required | default
------------|---------|------------|----------
hasHalo | bool | no | false

But the above implementation is a bit confusing, because the prop has no value but defaults to true

src/components/Avatar/Avatar.test.js
mattpauldavies: @jfrej and I can take a look at this next :)

src/components/Avatar/Avatar.story.js
mattpauldavies: We absolutely could use `image-loader` for images, this is ONLY here because we wanted to explicitly test an data URI image (as a one-off).  With that in mind would you still like this to be moved?

src/components/Avatar/AvatarCard.js
mattpauldavies: Love it, will adjust

src/components/Avatar/Avatar.scss
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        &::before {
+            position: absolute;
+            top: -2px;
+            left: -2px;
+            z-index: $halo-z-index;
+            display: block;
+            content: "";
+            background-image: url("avatar-halo.png");
```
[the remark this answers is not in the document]
mattpauldavies: It was and that is what I did initially.

However, the CSS gradient proved to be very jerky when animated in IE...

See: https://mdshowto.atlassian.net/browse/MDU-67

The jerkiness is caused by IE not handling border-radius very well.

src/components/Avatar/AvatarCard.js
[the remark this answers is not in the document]
mattpauldavies: Good spot, `--has-profile` used to be used, I refactored and removed...
I will remove the class altogether

src/components/Tooltip/TooltipBox.js
mattpauldavies: Yup! 👍

src/storybook-addons/storyWrapper.test.js
mattpauldavies: Very good point, I've added some tests to be thorough :)

src/components/Avatar/Avatar.story.js
mattpauldavies: This is the React preferred implementation (and in fact enforced by our linter).  It's based on the HTML spec...

`hasHalo` without assignation is not `null` it is `true`

The following would be no value: *(and invalid, per the prop-type validation)*

```
<Avatar hasHalo="" ... />
<Avatar hasHalo={null} ... />
```

This is valid for turning off the halo:

```
<Avatar ... /> // uses default which is false
<Avatar hasHalo={false} ... />
```

This is valid for turning on the halo:  *(these are equivalent)*

```
<Avatar hasHalo ... />
<Avatar hasHalo={true} ... />
```

 I hope that makes sense!

src/components/Avatar/Avatar.story.js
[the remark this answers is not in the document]
jfrej: For the record: the alignment is still wrong on my machine but we decided to leave it as is. Emojis are rendered differently on different systems and it's impossible to fix it everywhere. We should supply SVG icons to be used as avatar status icons if we want a consistent behaviour.

src/components/About.story.js
```diff
@@ -0,0 +1,78 @@
...
+### Usage
+
+You can consume UI React in one of two ways, as a whole library or as individual components.
+
+    import { YourComponent } from '@mudano/ui-react'; // UMD Modules
+
+    import { YourComponent } from 'node_modules/@mudano/ui-react/es/index'; // Native ES6 Modules
+
+    import YourComponent from 'node_modules/@mudano/ui-react/es/components/YourComponent/YourComponent';
+
```
jfrej: Were you still going to make changes to this section?

mattpauldavies: I think we should... I'll do that now

src/components/Avatar/Avatar.story.js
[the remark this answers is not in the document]
mattpauldavies: To follow up on this issue, we have agreed with Antoine and Makis to remove emoji support in favour of SVG icons.  Later we plan to explore a better way to support emojis

src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,102 @@
...
+const propTypes = {
+    className: PropTypes.string,
+    hasHalo: PropTypes.bool,
+    icon: PropTypes.element,
+    initials: InitialsPropType,
+    name: PropTypes.string.isRequired,
+    onClick: PropTypes.func,
+    size: ListPropType(['xs', 'sm', 'md', 'lg']),
+    src: PropTypes.string,
+    style: PropTypes.objectOf(PropTypes.oneOfType([
```
andyberry88: Can we abstract this out in to a `StyleObjectPropType`?

andyberry88:
Looks good. One comment re. refactoring out a style prop type.

PR #27, opened 2018-01-15
New automation workflow config

andyberry88:
- adds an empty `CODEOWNERS` file which GitHub can use to enforce PR approval if core code is changed
- adds a pull request template which devs can follow when manually creating a PR
- adds a commit message template - this can be applied by running `git config --local commit.template COMMIT_MSG`

The `CODEOWNERS`, `PULL_REQUEST_TEMPLATE` and `COMMIT_MSG` files are intended to evolve with the project - e.g. if the PR template doesn't contain the right information, or it's too heavyweight, it can be changed.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,26 +1,15 @@
-<!-- Remember to use a meaningful title that references the story ID (or story IDs) that this PR fixes. For example: [AB-123] Adds some cool new feature -->
+<!-- Remember to use a meaningful title which can be used in release notes. For example: Adds some cool new feature -->
 
-* **What kind of change does this PR introduce?**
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,26 @@
+<!-- Remember to use a meaningful title that references the story ID (or story IDs) that this PR fixes. For example: [AB-123] Adds some cool new feature -->
+
+* **What kind of change does this PR introduce?**
```
mattpauldavies: Did you want to append the pull request description to the changelog when we release?

If so is this the format that you want?  Or did you want to make use of titles instead of questions for each section so that it makes more sense for the person reading the changelog...

See #24 for example

PR #44, opened 2018-01-19
UMD Bundle and CommonJS Modules

mattpauldavies:

_none_

**New Features**

 - Adds UMD bundle, accessible as the primary entry point for the package
 - Adds CommonJS modules with individual CSS as a third option (keeps ES6 native modules)

**Bug Fixes**

- Fixes bug that prevented access to Avatar and Tooltip components

**Miscellaneous** 

- Updated usage documentation


lib/index.js
```diff
@@ -1,4 +1,14 @@
+import Avatar from '../src/components/Avatar/Avatar';
```
andyberry88: Can we import these from lib so there's one less thing to remember to add when creating a new component?

jfrej: I'm starting to think these should be automatically generated by the build script. I've seen other component libraries do it.

package.json
```diff
@@ -13,6 +14,14 @@
     "url": "git+https://github.com/Mudano/ui-react.git"
   },
   "author": "Mudano Ltd. (http://mudano.com)",
+  "contributors": [
+    "Andy Berry <andy.berry@mudano.com",
+    "Antoine Fromentin <antoine.fromentin@mudano.com>",
+    "David Corbacho <david.corbacho@mudano.com>",
+    "Jakub Frej <jakub.frej@mudano.com",
+    "Makis Spyratos <makis.spyratos@mudano.com>",
+    "Matt Davies <matt.davies@mudano.com>"
```
jfrej: Should these be our github-associated emails? Or perhaps `@users.noreply.github.com` to prevent spam like many people here choose to do https://github.com/facebook/react/blob/master/AUTHORS ?

mattpauldavies: Perhaps lose the emails altogether?

lib/index.js
mattpauldavies: I think automatically generated would be good, we can look at that later though...

I will swap to lib now...

package.json
andyberry88: Or just put `"The Mudano UI React Team": "info@mudano.com"`

mattpauldavies: Chatted with Andy on Slack, will remove

lib/index.js
```diff
@@ -1,4 +1,14 @@
...
+import TooltipBox from './TooltipBox';
 
-console.error('You are attempting to use UI React incorrectly, please check the documentation!'); // eslint-disable-line no-console
+export {
+    Avatar,
+    AvatarCard,
+    AvatarTitle,
+    Tooltip,
+    TooltipBox,
+};
```
jfrej: We should add Button and IconArrow here

PR #46, opened 2018-01-25
Replace istanbul-instrumenter-loader with babel-plugin-istanbul

jfrej:
_None_

**New Features**
_None_

**Bug Fixes**
_None_

**Miscellaneous**

- Replaced istanbul-instrumenter-loader with babel-plugin-istanbul


jfrej:
More info:
The babel plugin handles code instrumentation better and in particular doesn't cause wrong coverage reports for classes with a constructor and call to super()
See issue 106 of remap-istanbul

PR #47, opened 2018-01-26
TextField and TextArea Component

mattpauldavies:

_None_

**New Features**

New components:

- TextField
- TextArea
- IconClear
- IconNotification
- IconRequired
- IconSearch

**Bug Fixes**

- Fixed some minor rendering issues with Icons

**Miscellaneous**

- Clarify that issues should be reported via email (ui-react@mudano.com)


src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+    style: null,
+    tooltipError: null,
+    tooltipHint: null,
+    tooltipRequired: 'required',
+    value: '',
+};
+
+let lastInstanceId = 0;
+
+class textarea extends Component {
```
jfrej: Is `textarea` here intentionally all lower case? Normally we name classes using upper CamelCase.

src/components/TextField/TextField.js
```diff
@@ -0,0 +1,317 @@
...
+                        'uir-textfield--has-right-icon': this.props.isRequired || this.props.isClearable,
+                        'uir-textfield--has-value': this.state.value,
+                        'uir-textfield--invalid': this.props.isValid === false,
+                        'uir-textfield--readonly': this.props.isReadOnly,
+                        'uir-textfield--title': this.props.variant === TextFieldVariant.TITLE,
+                        'uir-textfield--valid': this.props.isValid,
+                    },
+                    this.props.className,
+                )}
+                id={this.props.id}
```
jfrej: I'm not sure if I agree with this. When providing a TextField `id`, I would expect it to be applied to the actual field - the input element.
Also, this reminds me: we need a `name` prop too to control the `name` attribute of the input.

src/components/Textfield/Textfield.scss
```diff
@@ -0,0 +1,169 @@
+@import "../variables";
```
jfrej: I'm on a case insensitive file system and on my machine this file overwrites the TextField/TextField.scss, so that I only have the incorrect version and the compilation breaks.
You probably have to use `git mv` to resolve this: https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git

package.json
```diff
@@ -15,10 +15,11 @@
   },
   "author": "Mudano Ltd. (http://mudano.com)",
   "license": "Apache-2.0",
+  "homepage": "https://mudano.github.io/ui-react/",
   "bugs": {
-    "url": "https://github.com/Mudano/ui-react/issues"
+    "url": "https://mdshowto.atlassian.net/projects/UR/issues",
+    "email": "techops@mudano.com"
```
jfrej: Shouldn't this be the same email address as in the README (`ui-react@mudano.com`)?

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+        PropTypes.string,
+    ]),
+    value: PropTypes.string,
+};
+
+const defaultProps = {
+    className: null,
+    componentRef: null,
+    hasAutoHeight: TextAreaAutoHeight.DISABLED,
+    hasLabelAlways: false,
```
jfrej: I thought that Antoine wanted this to be true by default. But perhaps I misunderstood, we should double check.
If it's always visible then the prop should have a different name.

jfrej:
Also, looks like you forgot to add your components to `/lib`

src/components/TextField/TextArea.js
mattpauldavies: Ahh, thank you for flagging. This was a mistake - will rectify

src/components/TextField/TextField.js
mattpauldavies: Question is, why do we have the `id` prop?

If it's to allow the user to identify (and then manipulate) the component, then an `id` on the wrapper does work.  If the user needs the input they can do the equivilant of `$('#compid  input')`

If on the other hand, the `id` is to identify the input then perhaps `id` should be linked to the input, not the wrapper, however, we give the `input` an automatically assigned `id` to allow the label to be linked via `for`

On balance I think I would leave it as it is ;)

package.json
mattpauldavies: well spotted! thank you

src/components/TextField/TextArea.js
mattpauldavies: Very good point... will check

src/components/TextField/TextArea.scss
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    &--invalid:after {
+        background: linear-gradient(to right, $pink 0%, transparent 100%);
+    }
+
+    // @NB: removed at Antoine's request, this may be restored later
+    // &--valid:after {
+    //     background: linear-gradient(to right, $teal 0%, transparent 100%);
+    // }
```
jfrej: I think we should just remove this. If it's needed we can recover this from git history.

src/components/TextField/TextArea.test.js
```diff
@@ -0,0 +1,255 @@
...
+
+    it('updates the component state when changing the textarea value', () => {
+        const textArea = shallow(<TextArea />);
+        const instance = textArea.instance();
+        instance.inputRef = { value: 'test' };
+        const mock = sinon.mock(instance);
+        const expectation = mock.expects('setState');
+        textArea.find('textarea').simulate('change');
+        expect(expectation).to.be.called();
+        mock.restore();
```
jfrej: Btw, this should also work:

    const setStateSpy = sinon.spy(textArea.instance(), 'setState');
    ...
    expect(setStateSpy).to.have.been.called();

src/components/TextField/TextField.js
```diff
@@ -0,0 +1,317 @@
...
+
+    handleInputFocus = (event) => {
+        const { onFocus } = this.props;
+        this.setState({
+            hasFocus: true,
+            showTooltip: true,
+        });
+        if (onFocus) {
+            onFocus(event);
+        }
```
jfrej: Same as my comments on TextArea, I think we should use `setState` callback here.

src/components/TextField/TextField.scss
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    &--invalid:after {
+        background: linear-gradient(to right, $pink 0%, transparent 100%);
+    }
+
+    // @NB: removed at Antoine's request, this may be restored later
+    // &--valid:after {
+    //     background: linear-gradient(to right, $teal 0%, transparent 100%);
+    // }
```
jfrej: Same as my comment on the TextArea, I think we should remove this until it's needed.

src/components/TextField/TextField.story.js
```diff
@@ -0,0 +1,223 @@
...
+import { TextFieldVariant } from './TextFieldEnums';
+import IconSearch from '../Icon/IconSearch';
+import IconNotification from '../Icon/IconNotification';
+
+const stories = storiesOf('TextField.TextField', module);
+
+stories.add(
+    'Overview',
+    storyWrapper(
+        'TextField is text input field.',
```
jfrej: Should this be "TextField is **a** text input field."?

src/components/TextField/TextFieldEnums.js
```diff
@@ -0,0 +1,11 @@
...
+export const TextFieldVariant = Object.freeze({
+    DEFAULT: 'default',
+    TITLE: 'title',
+});
+
+// TextAreaAutoHeight enum used by the TextArea autoHeight prop
+export const TextAreaAutoHeight = Object.freeze({
+    ENABLED: true,
+    DISABLED: false,
+});
```
jfrej: As per my comment in the TextArea, I think we don't need this enum, we should just use a boolean.

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+import TooltipBox from '../Tooltip/TooltipBox';
+import { TooltipBoxStatus, TooltipPosition } from '../Tooltip/TooltipEnums';
+import IconRequired from '../Icon/IconRequired';
+import { TextAreaAutoHeight } from './TextFieldEnums';
+import './TextArea.scss';
+
+const propTypes = {
+    className: PropTypes.string,
+    componentRef: PropTypes.func,
+    hasAutoHeight: ListPropType([
```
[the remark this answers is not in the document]
mattpauldavies: Fair enough, perhaps trying to be too clever here.  I'll remove the enums and go for a boolean

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+
+    handleInputBlur = (event) => {
+        const { onBlur } = this.props;
+        this.setState({
+            hasFocus: false,
+            showTooltip: false,
+        });
+        if (onBlur) {
+            onBlur(event);
+        }
```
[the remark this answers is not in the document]
mattpauldavies: I'm not 100% sure about this, I see your point and it is safer, however, does it matter if the `onBlur` function is called in parallel with `setState`?

I can't see a scenario where that would be a problem... with that in mind I like the idea of keeping the components as fast as possible

@andyberry88 can I get your opinion on this?

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    handleInputKeyUp = (event) => {
+        const { onKeyUp } = this.props;
+        if (onKeyUp) {
+            onKeyUp(event.key, event);
+        }
+    }
+
+    wrapInputWithTooltip = (input, tooltip) => {
```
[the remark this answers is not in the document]
mattpauldavies: Good idea

src/components/TextField/TextArea.scss
mattpauldavies: Agreed!

src/components/TextField/TextField.scss
mattpauldavies: Removed, same as TextArea

src/components/TextField/TextField.story.js
mattpauldavies: Thank you – well spotted

src/components/TextField/TextFieldEnums.js
mattpauldavies: Sorted

src/components/TextField/TextArea.js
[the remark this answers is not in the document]
jfrej: The problem is that they won't happen in parallel. It's very likely that `onBlur` will fire before state has been updated. Just search for "setState callback" and you'll find plenty of people having problems when not using callbacks.
This might be less of a problem with blur (although we can't imagine all the possible use cases) but is very likely with `onChange` - the `onChange` handler could fire before the internal `value` has been updated. This could lead to some weird behaviours. Even if current component code wouldn't have a problem with this discrepancy, managing state this way may cause issues in the future (there were bugs caused by this in m-dash-ui).

I think it's good practice to ensure internal state is updated before calling external functions.

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+                        'uir-textarea--valid': this.props.isValid,
+                    },
+                    this.props.className,
+                )}
+                onMouseEnter={this.handleMouseEnter}
+                onMouseLeave={this.handleMouseLeave}
+                style={this.props.style}
+            >
+                <div className="uir-textarea-inner">
+                    <div className="uir-textarea-label-wrapper">
```
jfrej: Should this <div> be rendered when `label` is `null`?

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+                        {label}
+                    </div>
+                    {this.wrapInputWithTooltip(
+                        <textarea
+                            aria-invalid={this.props.isValid === false}
+                            aria-required={this.props.isRequired}
```
jfrej: MDN says this:

> Now that required is available to more than 97% of users globally, it is no longer recommended that you use both required and aria-required. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints#Required_and_invalid_fields

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,279 @@
...
+                    }
+                >
+                    {textarea}
+                </Tooltip> :
+                textarea
+        );
+    }
+
+    render() {
+        /* eslint-disable jsx-a11y/label-has-for */
```
jfrej: Should we move this line closer to the actual label code instead of wrapping the whole `render` function?

src/components/TextField/TextArea.js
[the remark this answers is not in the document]
andyberry88: I agree with @jfrej on this. We should make sure the internal state has been set before triggering events etc. outside of the component.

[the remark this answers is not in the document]
mattpauldavies: Ok chaps, I bow to your expertise... I'll get it changed :)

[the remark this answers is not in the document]
andyberry88: https://media.giphy.com/media/SKjTHluaRvLG/giphy.gif

package.json
```diff
@@ -15,10 +15,11 @@
   },
   "author": "Mudano Ltd. (http://mudano.com)",
   "license": "Apache-2.0",
+  "homepage": "https://mudano.github.io/ui-react/",
   "bugs": {
-    "url": "https://github.com/Mudano/ui-react/issues"
+    "url": "https://mdshowto.atlassian.net/projects/UR/issues",
```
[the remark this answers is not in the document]
mattpauldavies: My thought process was as this is mainly for internal devs and we don't want to use GitHub then we should probably just point to the internal `url`

The other option is of course to remove, I don't believe `url` is required – it just adds a link on NPM which is kind of cool

Want to remove?

src/components/TextField/TextArea.test.js
mattpauldavies: Nice 👍

src/components/TextField/TextArea.js
mattpauldavies: good call, have updated

src/components/TextField/TextArea.js
mattpauldavies: Thanks for spotting this, originally label was inside the JSX so I couldn't place the comment alongside, then when I moved the label out of the JSX I didn't move the comment – good suggestion

src/components/TextField/TextArea.js
mattpauldavies: yeah I know this is the most elegant, it's there to prevent the label appearing and causing the UI to nudge up and down... I think it needs to be kept :)

package.json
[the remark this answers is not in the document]
andyberry88: Why not add a page to Storybook re. internal vs external usage + bugs and link to that?

bin/publish-release
```diff
@@ -49,32 +56,45 @@ const { getCurrentBranch, pullLatest, checkoutBranch } = require('../release-src
...
+    return 0;
+}
 
-return 0;
+try {
+    return publish();
+} catch (error) {
+    console.error(error);
+    return 1;
+}
```
jimmed: Why are we not using [ink](http://npm.im/ink) for this?

lib/IconClear.js
```diff
@@ -0,0 +1,3 @@
+import IconClear from '../src/components/Icon/IconClear';
+
+export default IconClear;
```
jimmed: We should consider using [babel-plugin-transform-export-extensions](https://babeljs.io/docs/plugins/transform-export-extensions/), which would allow this syntax instead:

```js
export default from '../src/components/...'
```

lib/index.js
```diff
@@ -5,6 +5,13 @@ import AvatarTitle from './AvatarTitle';
 import Button from './Button';
 
 import IconArrow from './IconArrow';
+import IconClear from './IconClear';
+import IconNotification from './IconNotification';
+import IconRequired from './IconRequired';
+import IconSearch from './IconSearch';
+
+import TextField from './TextField';
+import TextArea from './TextArea';
```
jimmed: Again, using extended exports:

```js
export { default as Button } from './Button'
export { default as IconArrow } from './IconArrow'
```

etc.

jfrej: This is explicitly forbidden by the Airbnb JS code style, see https://github.com/airbnb/javascript#modules--no-export-from-import

If we want to use this style, we should amend our code styleguide to allow it.

src/components/TextField/TextField.scss
```diff
@@ -0,0 +1,164 @@
...
+            clip: rect(1px, 1px, 1px, 1px);
+        }
+    }
+
+    &-input {
+        width: 100%;
+        padding: 2px 0 3px;
+        font-family: inherit;
+        font-size: inherit;
+        border: 0;
```
jfrej: You removed `outline: none` from here in this commit: 5cce8464d43bb157157e3d130eff698b3234b13c

Are you sure this is correct? Now there's a massive blue highlight on focus and we have the border underline extending on focus so I don't think we need the outline.

Not sure if `outline: none` should be added here or to `:focus`

src/components/TextField/TextField.js
```diff
@@ -0,0 +1,319 @@
...
+                    htmlFor={this.uid}
+                    className={cx(
+                        'uir-textfield-label',
+                        {
+                            'uir-textfield-label--visually-hidden': this.props.variant === TextFieldVariant.TITLE,
+                        },
+                    )}
+                >
+                    {this.props.label}
+                    {requiredIcon}
```
jfrej: Because this got moved into the `label`, we now don't show the required icon for title fields. I think that's wrong. If label is not visible, the required icon should still be rendered.

jfrej:
In IE11 clearable field has two clear icons. Can we fix this somehow?

src/components/Icon/IconClear.js
```diff
@@ -0,0 +1,42 @@
...
+
+let lastInstanceId = 0;
+
+const propTypes = {
+    className: PropTypes.string,
+    style: StyleObjectPropType(),
+};
+
+const defaultProps = {
+    className: '',
```
jfrej: We should probably use `null` as the default for consistency with the main components. I've already updated IconArrow and IconMore in my PR.

lib/IconClear.js
jimmed: Disregard, as per [comment below](https://github.com/Mudano/ui-react/pull/47#discussion_r166582945)

src/components/Icon/IconArrow.test.js
```diff
@@ -0,0 +1,10 @@
+/* global describe, expect, it, shallow */
+import React from 'react';
+import IconArrow from './IconArrow';
+
+describe('IconArrow', () => {
+    it('has the appropriate classes', () => {
```
andyberry88: Can we have a test that validates each icon renders SVG? I don't think we need to verify much (if any) of the SVG content but it would be useful to have a test to prove the component actually renders.

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    handleInputBlur = (event) => {
+        const onBlur = this.props.onBlur || (() => {});
+        this.setState({
+            hasFocus: false,
+            showTooltip: false,
+        }, () => {
+            onBlur(event);
```
andyberry88: We need to handle the `onBlur` (or other callbacks) throwing an error. There is no defined behaviour for this.

src/components/TextField/TextArea.js
```diff
@@ -0,0 +1,276 @@
...
+            showTooltip: true,
+        }, () => {
+            onFocus(event);
+        });
+    }
+
+    handleInputKeyDown = (event) => {
+        const { onKeyDown, onEnterKey } = this.props;
+        if (onKeyDown) {
+            onKeyDown(event.key, event);
```
andyberry88: Handling errors from events is more important here. If `onKeyDown` errors, `onEnterKey` never gets called. That might be what we want, but it isn't obvious or documented from what I can see.

src/components/TextField/TextField.js
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+
+    handleInputBlur = (event) => {
+        const onBlur = this.props.onBlur || (() => {});
+        this.setState({
+            hasFocus: false,
+            showTooltip: false,
+        }, () => {
+            onBlur(event);
```
andyberry88: Same thing re. handling errors from events.

bin/publish-release
mattpauldavies: Because we didn't know Ink existed!!!  And ooooh useful!

This originally started as a bash script, I then converted it into a JS script – we didn't want to spend too much time on it

If we get the opportunity to come back and improve we'll definitely look at Ink.

Thanks for the tip!

src/components/TextField/TextArea.js
mattpauldavies: Can we discuss this please?

src/components/TextField/TextField.scss
mattpauldavies: Well spotted, thank you.  Fixed and refined in both TextField and TextArea

src/components/Fields/RequiredIconAndTooltip.scss
```diff
@@ -0,0 +1,9 @@
+.uir-textfield-inner,
+.uir-textarea-inner {
+
+    & > .icon-required-tooltip {
+        position: absolute;
+        top: 20px;
+        width: auto !important;
```
jfrej: Could you please add a comment in code explaining why we need to use `!important`?
Generally we should avoid it unless there's a super good reason to keep it.

mattpauldavies: I know we should avoid, I've used it here because the `icon-required-tooltip' clashes with the standard TextField or TextArea tooltip... I'll see if I can refactor to remove and if not I'll add a comment

mattpauldavies: Found a better solution by adjusting the CSS classes 👍

src/components/About.story.js
```diff
@@ -171,3 +181,40 @@ Good luck!
...
+
+### Class Names
+
+Every component class name takes the format \`uir-component-name\` e.g. \`uir-avatar\`.
+
+When modifying a component we use the format \`uir-component--modifier\` e.g. \`uir-avatar--halo\`.
+
+### Callbacks and Errors
+
+Event callbacks provided to components should be called after the components internal state has been updated.
```
jfrej: Should this have an apostrophe after _components_, like so:
"after the components' internal state has been updated"?

Or perhaps use singular:
"after the component internal state has been updated."

src/components/Fields/TextArea.story.js
```diff
@@ -0,0 +1,159 @@
...
+            <TextArea label="Input with a hint" tooltipHint="My Example Hint" />
+        </div>,
+    ),
+);
+
+stories.add(
+    'Required',
+    storyWrapper(
+        'isRequired allows you to mark an input as required. And adds required icon with tooltip.',
+        <TextArea label="Required input" isRequired>
```
jfrej: I think it'd be good to add `tooltipRequired` to this example because it's otherwise undocumented (the source code for the variant below is not visible in storybook).

src/components/Fields/TextArea.test.js
```diff
@@ -0,0 +1,253 @@
...
+import TextArea from './TextArea';
+import Tooltip from '../Tooltip/Tooltip';
+import IconRequired from '../Icon/IconRequired';
+
+function createTextarea() {
+    return shallow(<TextArea label="My Input" />);
+}
+
+describe('TextArea', () => {
+    it('renders an div wrapper element', () => {
```
jfrej: Could you please use the new `isConformant` helper in this file to ensure correct classes, etc.?

src/components/Fields/TextField.story.js
```diff
@@ -0,0 +1,235 @@
...
+
+stories.add(
+    'Required',
+    storyWrapper(
+        'isRequired allows you to mark an input as required. And adds required icon with tooltip.',
+        <TextField
+            label="Required input"
+            value="Hello, Friend"
+            isRequired
+        />,
```
jfrej: As with `TextArea`, I think it'd be good to add `tooltipRequired` to this example because it's otherwise undocumented.

src/components/Fields/TextField.test.js
```diff
@@ -0,0 +1,308 @@
...
+import IconSearch from '../Icon/IconSearch';
+import IconRequired from '../Icon/IconRequired';
+import Button from '../Button/Button';
+
+function createTextfield() {
+    return shallow(<TextField label="My Input" />);
+}
+
+describe('TextField', () => {
+    it('renders an div wrapper element', () => {
```
jfrej: Could you please use the new isConformant helper in this file to ensure correct classes, etc.?

PR #52, opened 2018-02-01
Tabs Component

jfrej:

_None_

**New Features**

- `Tabs` component
- `IconMore` component

**Bug Fixes**

_None_

**Miscellaneous**

- Updated `Button` component colours


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,72 @@
...
+import StyleObjectPropType from '../../prop-types/style';
+import TabsNav from './TabsNav';
+import TabsPanes from './TabsPanes';
+import TabsPane from './TabsPane';
+import TabsTab from './TabsTab';
+
+let lastInstanceId = 0;
+
+const propTypes = {
+    selectedIndex: PropTypes.number,
```
mattpauldavies: Could we sort this prop list alphabetically like the other components please :)

src/components/_variables.scss
```diff
@@ -8,6 +8,7 @@ $font-family: "DINOT", sans-serif !default;
 $white: #fff;
 $grey: #ccc !default;
 $charcoal: #3a333c !default;
```
mattpauldavies: Let's check with Antoine although I think `charcoal` could be replaced with `slate` in all components

mattpauldavies:
Overall, the code looks really good, just a couple of little comments above...

I checked in Firefox and there are some odd issues with the drop-down:

screen shot 2018-02-01 at 14 38 37

src/components/Tabs/Tabs.js
jfrej: Oh, fixed!

mattpauldavies:
Apart from agreeing which version of grey we want to use this all looks excellent.

Just one question: did you want to implement that trick you found to stop the menu nudging as it swaps between bold and normal?

Could we add that as an improvement to come back to later?

src/components/Tabs/TabsPane.js
```diff
@@ -0,0 +1,87 @@
...
+    relatedTabId: null,
+};
+
+class TabsPane extends Component {
+    state = {
+        isDeselecting: false,
+        isSelecting: false,
+    };
+
+    componentWillReceiveProps(nextProps) {
```
mattpauldavies: `componentWillReceiveProps` is a little verbose... could this work as a more concise option:
```
componentWillReceiveProps(nextProps) {
    this.setState({
        isDeselecting: this.props.isSelected && !nextProps.isSelected,
        isSelecting: !this.props.isSelected && nextProps.isSelected,
    });
}
```

jfrej: oh, yes, good point!

jfrej: Fixed

src/components/Icon/IconMore.test.js
```diff
@@ -0,0 +1,10 @@
+/* global describe, expect, it, shallow */
+import React from 'react';
+import IconArrow from './IconMore';
+
+describe('IconMore', () => {
+    it('has the appropriate classes', () => {
```
andyberry88: Can we have a test to verify `render()` returns SVG content? (see https://github.com/Mudano/ui-react/pull/47#discussion_r166914332). We probably need to add these tests for each icon component.

jfrej: Test added.
I've also created a helper for that which I'll include in a separate PR.

andyberry88: 👍

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,75 @@
...
+
+        lastInstanceId += 1;
+
+        this.state = {
+            selectedIndex: this.props.selectedIndex,
+        };
+    }
+
+    render() {
+        const children = React.Children.map(this.props.children, (child) => {
```
mattpauldavies: I wonder if we should split this into multiple functions to make it easier to read and maintain?
Perhaps:
`getChildren()`
Which calls:
`getTabsNav()`
`getTabsPane()`

jfrej: Changed

PR #55, opened 2018-02-08
Button aria-expanded prop

jfrej:

_None_

**New Features**

- Added `aria-expanded` prop to the `Button` component

**Bug Fixes**

_None_


jfrej:
There are two commits here that you might want to review separately:

1. Adds the new prop and tests
2. Refactors Button's render() function to use `this.props.` for consistency with other components

PR #56, opened 2018-02-09
Unit test helpers

jfrej:
Let's not include this one in the changelog.

_None_

**New Features**

_None_

**Bug Fixes**

- Fixed `AvatarTitle` component function name. It doesn't affect the API but fixes the storybook docs.

**Miscellaneous**

- Added unit test helpers for testing component HTML tag and component className.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "karma-phantomjs-launcher": "^1.0.4",
     "karma-spec-reporter": "^0.0.32",
     "karma-webpack": "^2.0.8",
+    "lodash": "^4.17.5",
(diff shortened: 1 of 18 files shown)


jfrej:
@andyberry88 I added a single `isConformant` test helper to run the most common tests:
- checks the rendered HTML tag (assumes `<div>` by default)
- checks if component `className` is correct and matches our naming convention (`uir-` prefix plus component name in kebab case)
- checks if `className` prop is implemented
- checks if `style` prop is implemented
- optionally checks if component can render children of any type (string, node)

PR #57, opened 2018-02-13
DateField and InlineDatePicker

mattpauldavies:

_None_

**New Features**

 - DateField
 - DateFieldLinked
 - InlineDatePicker

**Bug Fixes**

_None_


jfrej:
It's looking good. But there are some issues with keyboard support. We need to make sure that the datepickers are fully accessible.
Have a look at this example: http://whatsock.com/tsg/Coding%20Arena/ARIA%20Date%20Pickers/ARIA%20Date%20Picker%20(Basic)/demo.htm

And also these two articles:

- https://axesslab.com/accessible-datepickers/
- http://www.webaxe.org/accessible-date-pickers/

In general we might be able to use flatpickr for that, it's got some accessibility built in (although there is an issue in their issue log about some accessibility problems). It's a shame flatpickr has `aria-label` for day fields which doesn't include day of week. But perhaps that can be configured, I haven't checked.

src/components/Fields/DateField.js
```diff
@@ -0,0 +1,165 @@
...
+
+    datePickerOpen() {
+        this.setState({
+            showDatePicker: true,
+        });
+    }
+
+    formatDate = (date) => {
+        if (!date) {
+            return date;
```
andyberry88: Might be better to return something explicit here rather - e.g. `return undefined`

src/components/Fields/DateField.js
```diff
@@ -0,0 +1,165 @@
...
+        }
+        const dateGroupsRegex = /\w{3} (\w{3}) (\d{2}) (\d{4}) .+/;
+        return date.toString().replace(dateGroupsRegex, '$2 $1 $3');
```
andyberry88: Can we use a date library for formatting so it's more explicit what you're trying to do?

mattpauldavies: Discussed a date library on the #devs channel... Jim and others are VERY keen to avoid a date library.

I picked Flatpickr (the underlying calendar) because it uses the Date object and nothing more – I wanted to keep the component light and allow the features lanes to pick their own date library (if they want one).

My theory was that we may already have a date library in ST* and I didn't want to go and add another.

With that said I agree that bit of regex is fairly minging... options:

1. add a date library
2. add a comment to explain
3. leave it as is

What would you like to do?

* I checked ST package.json apparently we're using `date-fns`, `datejs` and `react-dates` (includes `moment`)

src/components/Fields/DateField.js
mattpauldavies: Good call... I've refactored – will discuss my suggested changes offline 👍

src/components/Fields/DateField.js
andyberry88: I agree on the choice of avoiding a date library. Users of `ui-react` are then not bound by the choice of date library here + we don't have any real need for a date library. The bit I dislike is the magic of the regex.

Can we either:

- find a minimal date formatting library
- move the regex function into our own 'library' and add a few tests around the formatting

PR #58, opened 2018-02-13
Release v1.2.0

mattpauldavies:
Automatic release of v1.2.0


jfrej:
LGTM
(although we should agree on the writing style in our changelog, what should be formatted as `code`, etc.)

PR #60, opened 2018-02-19
TextArea auto resizes by default

mattpauldavies:

- `hasAutoHeight` prop removed from TextArea

**New Features**

- `hasFixedHeight` prop added to TextArea

**Bug Fixes**

_None_


src/components/Fields/TextArea.js
```diff
@@ -29,6 +28,7 @@ const propTypes = {
     onKeyPress: PropTypes.func,
     onKeyUp: PropTypes.func,
     placeholder: PropTypes.string,
+    preventAutoResize: PropTypes.bool,
```
jfrej: I like `preventAutoResize` prop name, but if we want to keep consistency with other bool props I would go for `hasFixedHeight`. Then we can have a matching CSS class too.

src/components/Fields/TextArea.story.js
```diff
@@ -53,21 +63,22 @@ _NB: label will not hide if there is no value._
...
-    'Auto Height',
+    'Fixed Height',
     storyWrapper(
-        'hasAutoHeight allows the textarea to resize with the content.',
+        `
+hasFixedHeight stops the textarea from resizing as the content changes.
+
+The example below has a fixed height of 200px.
+
+_NB: this mode also enables the default textarea resize function._
```
jfrej: This is no longer true, now that you hid the native resize handle.

src/components/Fields/TextArea.test.js
```diff
@@ -100,9 +100,14 @@ describe('TextArea', () => {
...
-        const textArea = shallow(<TextArea hasAutoHeight />);
-        expect(textArea).to.have.className('uir-text-area--has-auto-height');
+    it('has a auto resize class if hasFixedHeight is true', () => {
+        const textArea = shallow(<TextArea hasFixedHeight />);
+        expect(textArea).to.have.className('uir-text-area--has-fixed-height');
+    });
+
+    it('does not have a auto resize class if hasFixedHeight is false', () => {
+        const textArea = shallow(<TextArea hasFixedHeight={false} />);
+        expect(textArea).to.not.have.className('uir-text-area--has-fixed-height');
```
jfrej: Both of these tests can now be removed since you no longer use the `--had-fixed-height` class.
It would be nice to have another test because effectively we're no longer testing the `hasFixedHeight` prop. But I'm not sure if it's possible to test if `autosize()` has been called or not.

src/components/Fields/TextArea.story.js
mattpauldavies: Thank you

PR #89, opened 2018-03-26
TextField icon click gives focus to input

mattpauldavies:

_None_

**New Features**

_None_

**Bug Fixes**

- TextField icon click gives input focus

### DIFF
--- a/src/components/Fields/TextField.js
+++ b/src/components/Fields/TextField.js
@@ -281,7 +281,8 @@ class TextField extends Component {
                     <Button
                         className="icon-wrapper"
                         icon={this.props.icon}
-                        onClick={this.handleClearIconClick}
+                        onClick={this.handleIconClick}
+                        tabIndex={-1}
                         variant={ButtonVariant.CLEAR}
                     />
                 </span>
--- a/src/components/Fields/TextField.js
+++ b/src/components/Fields/TextField.js
@@ -279,7 +279,7 @@ class TextField extends Component {
             (
                 <span className="uir-text-field-left-icon">
                     <Button
-                        className="icon-wrapper"
+                        className="uir-text-field-icon-wrapper"
                         icon={this.props.icon}
                         onClick={this.handleIconClick}
                         tabIndex={-1}
@@ -292,7 +292,7 @@ class TextField extends Component {
             (
                 <span className="uir-text-field-right-icon">
                     <Button
-                        className="icon-wrapper"
+                        className="uir-text-field-icon-wrapper"
                         icon={<IconClear />}
                         onClick={this.handleClearIconClick}
                         variant={ButtonVariant.CLEAR}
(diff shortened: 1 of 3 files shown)


src/components/Fields/TextField.js
```diff
@@ -265,19 +278,24 @@ class TextField extends Component {
         const icon = this.props.icon ?
             (
                 <span className="uir-text-field-left-icon">
-                    {this.props.icon}
+                    <Button
+                        className="icon-wrapper"
+                        icon={this.props.icon}
+                        onClick={this.handleClearIconClick}
```
jfrej: This shouldn't be here, right? Now when I click on any icon, the field clears.

src/components/Fields/TextField.js
```diff
@@ -265,19 +278,24 @@ class TextField extends Component {
...
+                        variant={ButtonVariant.CLEAR}
+                    />
                 </span>
             ) :
             null;
         const clearIcon = this.props.isClearable && this.state.value ?
             (
                 <span className="uir-text-field-right-icon">
                     <Button
+                        className="icon-wrapper"
```
jfrej: Should this class be prefixed with something?

src/components/Fields/TextField.scss
```diff
@@ -128,44 +128,36 @@ $textFieldLabelColor: $charcoal !default;
...
+    .uir-button.icon-wrapper {
+        min-width: 16px;
+        padding: 0;
+        margin: 0;
 
-            &:hover {
-                background-color: transparent;
-            }
+        &:hover {
+            background-color: transparent;
```
jfrej: This should also be added to `:focus` - currently there's a grey background on focus.

src/components/Fields/TextField.js
jfrej: Should we have a test that checks that the field doesn't clear ;) ?

src/components/Fields/TextField.scss
mattpauldavies: ahh thanks well spotted

src/components/Fields/TextField.js
mattpauldavies: possibly... I initially thought `uir-text-field-icon-wrapper` was a bit long!! but perhaps that's the best way to go

src/components/Fields/TextField.scss
```diff
@@ -128,44 +128,37 @@ $textFieldLabelColor: $charcoal !default;
...
+        &:focus,
+        &:hover {
+            background-color: transparent;
+        }
 
-            .uir-icon-clear:hover {
-                fill: $charcoal;
-            }
+        .uir-icon-clear:hover {
+            fill: $charcoal;
```
jfrej: It'd be nice to add `:focus` to this as well. Currently there is no indication of focus when I tab into the clear icon.

PR #90, opened 2018-03-26
TextField accepts zero as a valid input

mattpauldavies:

_None_

**New Features**

_None_

**Bug Fixes**

- allow zero to activate TextField and move label

### DIFF
--- a/src/components/Fields/TextField.test.js
+++ b/src/components/Fields/TextField.test.js
@@ -356,4 +356,19 @@ describe('TextField', () => {
         const textField = shallow(<TextField name={exampleName} />);
         expect(textField.find('input').prop('name')).to.equal(exampleName);
     });
+
+    it('adds --has-value when value is not null', () => {
+        const textField = shallow(<TextField value="test" />);
+        expect(textField.hasClass('uir-text-field--has-value')).to.equal(true);
+    });
+
+    it('adds --has-value when value is falsy but not null', () => {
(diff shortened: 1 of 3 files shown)


src/components/Fields/TextField.test.js
```diff
@@ -363,4 +363,19 @@ describe('TextField', () => {
...
+        const textField = shallow(<TextField value="test" />);
+        expect(textField.hasClass('uir-text-field--has-value')).to.equal(true);
+    });
+
+    it('adds --has-value when value is falsy but not null', () => {
+        const textField = shallow(<TextField value={0} />);
+        expect(textField.hasClass('uir-text-field--has-value')).to.equal(true);
+    });
+
+    it('does not add --has-value when value isnull', () => {
```
jfrej: Technically, this should say "when value is undefined". If you try setting the `value` to `null` it will actually add the `--has-value` class (but React will also log an error in the console about using `null`).

jfrej:
Looking good
Two questions:
1. Why do we need to suport number as a value? Why not just string? Is it to make it easier to use?
2. Do we need to do the same with TextArea?

mattpauldavies:
Good questions...
1) yes to make it easier (we allow Number or String as a value) for TextField
2) no TextArea only allows string for value

src/components/Fields/TextField.test.js
mattpauldavies: Ahh yes, good point

PR #92, opened 2018-03-26
Increase TextField font size from 13px to 16px

mattpauldavies:

- TextField font size increase from 13px to 16px

**New Features**

_None_

**Bug Fixes** 

_None_


mattpauldavies:
This should be released on it's own, to allow people to rollback if they increase in size breaks layouts

PR #96, opened 2018-03-27
Fix issue with TextField and TextArea not handling null

mattpauldavies:

_None_

**New Features**

_None_

**Bug Fixes**

- fix issue with TextField and TextArea not handling null

### DIFF
--- a/src/components/Fields/TextArea.story.js
+++ b/src/components/Fields/TextArea.story.js
@@ -40,6 +40,28 @@ stories.add(
             placeholder="your placeholder here"
             value="Hello, Friend"
         />,
+        <div>
+            <TextArea
+                label="Number Zero"
+                placeholder="your placeholder here"
+                value={0}
+            />
+            <TextArea
+                label="Empty String"
+                placeholder="your placeholder here"
+                value=""
+            />
+            <TextArea
+                label="Null Value"
+                placeholder="your placeholder here"
+                value={null}
+            />
+            <TextArea
+                label="Undefined Value"
+                placeholder="your placeholder here"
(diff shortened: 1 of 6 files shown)


src/components/Fields/TextArea.story.js
```diff
@@ -40,6 +40,28 @@ stories.add(
             placeholder="your placeholder here"
             value="Hello, Friend"
         />,
+        <div>
+            <TextArea
+                label="Number Zero"
+                placeholder="your placeholder here"
+                value={0}
```
jfrej: `TextArea` doesn't accept numbers, so this logs an error to the console.

mattpauldavies: valid point, I'll remove!

src/components/Fields/TextArea.test.js
```diff
@@ -277,4 +277,24 @@ describe('TextArea', () => {
...
+
+    it('does not add --has-value when value is null', () => {
+        const textField = shallow(<TextArea value={null} />);
+        expect(textField.hasClass('uir-text-area--has-value')).to.equal(false);
+    });
+
+    it('adds --has-value when value is falsy but not null', () => {
+        const textField = shallow(<TextArea value={0} />);
+        expect(textField.hasClass('uir-text-area--has-value')).to.equal(true);
+    });
```
jfrej: I'm not sure if we need this one if TextArea doesn't accept numbers

jfrej: I'm not sure if we need this one if TextArea doesn't accept numbers

PR #100, opened 2018-04-23
Breadcrumbs Component

mattpauldavies:

_None_

**New Features**

- new breadcrumb (with title) component

**Bug Fixes**

_None_


mattpauldavies:
This is ready for use, however, it is missing the animation that Antoine would like.  I'll endeavour to add animation before release although this could be added later...

PR #101, opened 2018-04-23
AvatarMenu Component

mattpauldavies:

_None_

**New Features**

- add AvatarMenu component

**Bug Fixes**

_None_


src/components/Avatar/AvatarMenu.js
```diff
@@ -0,0 +1,95 @@
...
+    children: PropTypes.arrayOf(PropTypes.element).isRequired,
+    onMenuItemClick: PropTypes.func,
+    open: PropTypes.bool,
+    position: PropTypes.string,
+    style: StyleObjectPropType,
+};
+
+const defaultProps = {
+    className: null,
+    onMenuItemClick: null,
```
Vazerthon: This could be set to `() => {}` which would remove the need for L41 I think

src/components/Avatar/AvatarMenu.test.js
```diff
@@ -0,0 +1,139 @@
...
+        // eslint-disable-next-line no-console
+        expect(console.error).to.be.called();
+    });
+
+    it('renders with Avatar', () => {
+        const avatarMenu = renderMockMenu();
+        expect(avatarMenu.find(Avatar).length).to.equal(1);
+    });
+
+    it('default open state is false', () => {
```
Vazerthon: This is quite picky but I'd prefer if this was 'sets open state to false by default', just so the whole thing reads as a sentence

Vazerthon:
Looks good to me (as far as I can tell at this stage)

src/components/Avatar/AvatarMenu.test.js
mattpauldavies: If you like mate, sounds good to me

src/components/Avatar/AvatarMenu.js
mattpauldavies: Agreed. Although on this front I'd been copying the style we've used throughout UI React.  @jfrej what do you think?  If we change to `() => {}` here I feel we should do it everywhere

jfrej: We used to do that and there was a reason we switched to using `null`, but I can't remember it at the moment.
I agree that if we changed it here, we should change it everywhere. So for now I would keep the `null` and consider this change in a separate PR.

mattpauldavies: Cool we'll stick with `null` for now then :)

src/components/Avatar/AvatarMenu.test.js
```diff
@@ -60,7 +60,7 @@ describe('AvatarMenu', () => {
         expect(avatarMenu.find(Avatar).length).to.equal(1);
     });
 
-    it('default open state is false', () => {
+    it('sets open state to false by default', () => {
```
Vazerthon: Thank you :-)

PR #103, opened 2018-04-23
Textfield prefix zero value bug

mattpauldavies:

_None_

**New Features**

_None_

**Bug Fixes**

- fix the issue with prefix not being displayed if Textfield has 0 as a value

### DIFF
--- a/src/components/Fields/TextField.js
+++ b/src/components/Fields/TextField.js
@@ -260,7 +260,7 @@ class TextField extends Component {
     }
 
     render() {
-        const hasValue = this.state.value === null ? false : `${this.state.value}`;
+        const hasValue = this.state.value === null ? false : Boolean(this.state.value);
         const showLabel = (
             this.props.autoHideLabel === false ||
             this.state.hasFocus ||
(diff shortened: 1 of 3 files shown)


src/components/Fields/TextField.js
```diff
@@ -260,11 +260,12 @@ class TextField extends Component {
     }
 
     render() {
+        const hasValue = this.state.value === null ? false : `${this.state.value}`;
```
jfrej: How about this instead:
`const hasValue = this.state.value === 0 ? true : Boolean(val)`
?

That way `hasValue` will always be a boolean like its name suggests instead of sometimes being a string.

```
Boolean(undefined) => false
Boolean(null) => false
Boolean('') => false
Boolean(' ') => true
Boolean('test') => true
Boolean('0') => true
```

src/components/Fields/TextField.js
```diff
@@ -260,11 +260,12 @@ class TextField extends Component {
     }
 
     render() {
+        const hasValue = this.state.value === 0 ? true : Boolean(this.state.value);
```
jfrej: Approved.

Although, now I'm thinking it would've been more elegant to just write this instead:
`const hasValue = this.state.value === 0 || Boolean(this.state.value);`

PR #104, opened 2018-04-24
New Icons

mattpauldavies:

_None_

**New Features**

- new dashboard icon
- new priority icon
- new project icon

**Bug Fixes**

_None_

### DIFF
--- a/src/components/Icon/Icon.scss
+++ b/src/components/Icon/Icon.scss
@@ -6,19 +6,14 @@
     fill: $charcoal;
 }
 
-.uir-icon-notification {
+.uir-icon-fill-pink {
     fill: $pink;
 }
 
-.uir-icon-required {
-    fill: $charcoal;
-}
-
-
-.uir-icon-arrow-fill {
+.uir-icon-fill-charcoal {
     fill: $charcoal
 }
 
-.uir-icon-arrow-stroke {
+.uir-icon-stroke-charcoal {
     stroke: $charcoal
 }
--- a/src/components/Icon/Icon.scss
+++ b/src/components/Icon/Icon.scss
@@ -3,17 +3,32 @@
 .uir-icon {
     width: 24px;
     height: 24px;
-    fill: $charcoal;
+
+    &-fill {
+        fill: $charcoal;
+    }
+
+    &-stroke {
+        stroke: $charcoal
+    }
 }
 
-.uir-icon-fill-pink {
+.uir-icon-path {
+
+    &-fill {
+        fill: $charcoal;
+    }
+
+    &-stroke {
+        stroke: $charcoal
+    }
+}
+
+.uir-icon-dashboard {
+    width: 32px;
+    height: 24px;
+}
+
+.uir-icon-notification {
     fill: $pink;
-}
-
-.uir-icon-fill-charcoal {
-    fill: $charcoal
-}
-
-.uir-icon-stroke-charcoal {
-    stroke: $charcoal
 }
--- a/src/components/Icon/IconDashboard.js
+++ b/src/components/Icon/IconDashboard.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import StyleObjectPropType from '../../prop-types/style';
+import './Icon.scss';
+
+let lastInstanceId = 0;
+
+const propTypes = {
+    className: PropTypes.string,
+    style: StyleObjectPropType,
+};
+
+const defaultProps = {
+    className: null,
+    style: null,
+};
+
+const IconDashboard = ({
+    className,
+    style,
+}) => {
+    lastInstanceId += 1;
+    const iconTitleId = `icon-dashboard-title-${lastInstanceId}`;
+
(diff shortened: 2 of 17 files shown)


src/components/Icon/IconProject.js
```diff
@@ -0,0 +1,44 @@
...
+    style,
+}) => {
+    lastInstanceId += 1;
+    const iconTitleId = `icon-project-title-${lastInstanceId}`;
+
+    return (
+        <svg
+            className={cx('uir-icon uir-icon-project', className)}
+            style={style}
+            viewBox="-4 -1 24 24"
```
jfrej: Do we have to have negative values in the `viewBox`? Looks weird

mattpauldavies: Might look weird but the icons are an odd shape, it's the best way I've found to get them to align correctly in the view box.  If we don't have the negative values it becomes a real pain to get the icons to align with text

src/components/Icon/IconDashboard.js
```diff
@@ -0,0 +1,48 @@
...
+            style={style}
+            viewBox="0 0 32 24"
+            aria-labelledby={iconTitleId}
+        >
+            <title id={iconTitleId}>Dashboard Icon</title>
+            <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+                <g fillRule="nonzero">
+                    <path className="uir-icon-fill-charcoal" fillOpacity="0.5" transform="translate(0,0)" d="M8 1.155L1.206 5.077v7.846L8 16.845l6.794-3.922V5.077L8 1.155zM8 0l7.794 4.5v9L8 18 .206 13.5v-9L8 0z" />
+                    <path className="uir-icon-fill-charcoal" fillOpacity="1" transform="translate(7.8,4.5)" d="M8 1.155L1.206 5.077v7.846L8 16.845l6.794-3.922V5.077L8 1.155zM8 0l7.794 4.5v9L8 18 .206 13.5v-9L8 0z" />
+                    <path className="uir-icon-fill-charcoal" fillOpacity="0.2" transform="translate(15.46,0)" d="M8 1.155L1.206 5.077v7.846L8 16.845l6.794-3.922V5.077L8 1.155zM8 0l7.794 4.5v9L8 18 .206 13.5v-9L8 0z" />
```
jfrej: Do we need this `uir-icon-fill-charcoal` class? It's not defined anywhere.
Where possible, classes on icon paths should describe the shapes to make it easier to style individual paths. See for example IconMore here: https://github.com/Mudano/ui-react/blob/master/src/components/Icon/IconMore.js#L34-L36
where classes are "uir-icon-more-oval-1"
If the colour changes, "-fill-charcoal" won't make any sense.

Btw, those icons aren't visible when I run the code locally.

src/components/Icon/IconPriority.js
```diff
@@ -0,0 +1,46 @@
...
+        <svg
+            className={cx('uir-icon uir-icon-priority', className)}
+            style={style}
+            viewBox="0 1 24 24"
+            aria-labelledby={iconTitleId}
+        >
+            <title id={iconTitleId}>Priority Icon</title>
+            <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+                <g fillRule="nonzero">
+                    <polygon className="uir-icon-fill-charcoal" points="12 19.5857864 16.7832395 14.802547 12.4460456 6.12815913 13 5 18 15 12 21 6 15 12 3 12.559017 4.11803399 7.21676051 14.802547" />
```
jfrej: Same question about the `uir-icon-fill-charcoal` class

src/components/Icon/IconDashboard.js
mattpauldavies: Pull the latest – it looks like when I pulled this code into it's own branch I lost the icon.scss file.

We use the same fill and stroke colour across multiple icons so I felt it made more sense to share those classes (see icon.scss)

src/components/Icon/IconPriority.js
mattpauldavies: see above

src/components/Icon/IconDashboard.js
jfrej: Well, I wholeheartedly disagree with having colour names in classes. The classes are supposed to help people style the icon, including a change of colour. So they should semantically describe the element, not how it looks. Because the look might change.

I don't want to block you so I can approve this, though.
But since you're changing existing class names, this will have to be a breaking change.

src/components/Icon/Icon.scss
```diff
@@ -3,22 +3,32 @@
...
-}
+    &-fill {
+        fill: $charcoal;
+    }
 
-.uir-icon-required {
-    fill: $charcoal;
+    &-stroke {
+        stroke: $charcoal
+    }
```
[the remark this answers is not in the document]
mattpauldavies: I left them in to provide defaults for future icons.  Want to remove?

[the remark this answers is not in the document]
jfrej: Yeah, I think that if they're not being used they should be removed until we need them.

PR #112, opened 2018-05-30
Fix AvatarMenu link width issue

mattpauldavies:

_None_

**New Features**

_None_

**Bug Fixes**

- fix the issue with links in AvatarMenu not filling full width of menu and being hard to click

### DIFF
--- a/src/components/Avatar/AvatarMenu.scss
+++ b/src/components/Avatar/AvatarMenu.scss
@@ -82,7 +82,7 @@ $avatar-menu-size: 52px;
             .uir-avatar-menu-item a:focus,
             .uir-avatar-menu-item a:active {
                 display: inline-block;
-                flex: 1;
+                flex: 1 0 0%;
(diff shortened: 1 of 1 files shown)


src/components/Avatar/AvatarMenu.scss
```diff
@@ -80,6 +82,7 @@ $avatar-menu-size: 52px;
             .uir-avatar-menu-item a:focus,
             .uir-avatar-menu-item a:active {
                 display: inline-block;
+                flex: 1;
```
Vazerthon: I _think_ IE11 needs this to be `flex: 1 0 0%`

Vazerthon: https://caniuse.com/#feat=flexbox it's mentioned somewhere in here

mattpauldavies: Good call - fixed!

PR #117, opened 2018-06-12
Add componentRef prop to Button

diogobeda:

_None_

**New Features** 

Adding `componentRef` prop to Button

**Bug Fixes** 

_None_


src/components/Button/Button.js
```diff
@@ -142,7 +144,11 @@ class Button extends Component {
     }
 
     handleRef = (ref) => {
-        this.componentRef = ref;
+        const { componentRef } = this.props;
+        this.buttonRef = ref;
```
diogobeda: `this.componentRef` wasn't use, so I took the opportunity to rename it to `buttonRef` to avoid any confusion

PR #138, opened 2018-10-31
Fixes TextField label overlapping the autofilled value in Chrome

jfrej:

_None_

**New Features** 

_None_

**Bug Fixes** 

- Fixes `TextField` label overlapping the autofilled field value in Chrome
Thanks to the Klarna team for the solution: https://medium.com/@brunn/detecting-autofilled-fields-in-javascript-aed598d25da7
Their implementation can be found here:
https://github.com/klarna/ui/blob/v4.10.0/Field/index.jsx#L104-L114
https://github.com/klarna/ui/blob/v4.10.0/Field/styles.scss#L228-L241

### DIFF
--- a/src/components/Fields/TextField.js
+++ b/src/components/Fields/TextField.js
@@ -107,12 +107,21 @@ class TextField extends Component {
         this.state = {
             hasFocus: false,
             hasMouseOver: false,
+            isAutofilled: false,
             showTooltip: false,
             value: props.value,
         };
 
         lastInstanceId += 1;
         this.uid = `text-field-${lastInstanceId}`;
+    }
+
+    componentDidMount() {
+        const { inputRef } = this;
+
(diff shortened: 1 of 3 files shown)


src/components/Fields/TextField.js
```diff
@@ -261,7 +287,8 @@ class TextField extends Component {
     }
 
     render() {
-        const hasValue = this.state.value === 0 ? true : Boolean(this.state.value);
+        const hasValue = this.state.isAutofilled ||
+            (this.state.value === 0 ? true : Boolean(this.state.value));
```
eibhleag: `x ? true : y` should reduce to `x || y` unless I'm missing something

jfrej: Ah, fair point. But I've already merged :/

PR #144, opened 2019-01-15
[SPS-29] Add autoComplete prop to TextField and TextArea

diogobeda:

_None_

**New Features** 

New `autoComplete` prop on TextField and TextArea

**Bug Fixes** 

_None_

**Performance Improvements** 

**Miscellaneous**

### DIFF
--- a/src/components/Fields/TextArea.js
+++ b/src/components/Fields/TextArea.js
@@ -23,6 +23,7 @@ const propTypes = {
     isValid: PropTypes.bool,
     label: PropTypes.string,
     name: PropTypes.string,
+    autoComplete: PropTypes.oneOf(['on', 'off']),
     onBlur: PropTypes.func,
     onChange: PropTypes.func,
     onEnterKey: PropTypes.func,
@@ -51,6 +52,7 @@ const defaultProps = {
     isValid: null,
     label: null,
     name: null,
+    autoComplete: null,
     onBlur: null,
     onChange: null,
     onEnterKey: null,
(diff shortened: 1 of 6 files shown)


src/components/Fields/TextField.story.js
```diff
@@ -400,3 +400,17 @@ _NB: if you add a prefix and an icon, only the prefix will be displayed_
         </div>,
     ),
 );
+
+stories.add(
+    'Autocomplete',
+    storyWrapper(
+        `
+\`autoComplete\` prop allows you to control whether the browser should autocomplete input value
+        `,
```
NikolaZavadska: screenshot 2019-01-15 at 13 24 05

looks like this `,` appears in storybook next to the field

diogobeda: oops :)

src/components/Fields/TextArea.js
```diff
@@ -51,6 +52,7 @@ const defaultProps = {
     isValid: null,
     label: null,
     name: null,
+    autoComplete: null,
```
williankrause: Wouldn't be good to have a default value assigned to it?

diogobeda: If we don't pass a default value it will just use browser's default, which is to have it on. So I don't see the need for us to add it.

jfrej: No, it's better to have no value which will make React not output the `autocomplete` attribute at all.
This will be valid HTML and consistent with the HTML spec - not having the `autocomplete` attribute means that the browser will look at the parent `<form>` tag to decide whether to use autocomplete or not.

src/components/Fields/TextField.js
```diff
@@ -30,6 +30,7 @@ const propTypes = {
     max: PropTypes.number,
     min: PropTypes.number,
     name: PropTypes.string,
+    autoComplete: PropTypes.oneOf(['on', 'off']),
```
jfrej: All props in ui-react should be ordered alphabetically. Please amend here and in the defaultProps definition.

PR #148, opened 2019-01-18
Improve IconProject SVG image alignment and add title prop

jfrej:

- Modified `IconProject`'s SVG image alignment and `viewBox` which might cause small layout issues

**New Features** 

- Added `title` prop to `IconProject`

**Bug Fixes** 

- `IconProject`'s `viewBox` no longer uses negative values

### DIFF
--- a/src/components/Icon/IconProject.test.js
+++ b/src/components/Icon/IconProject.test.js
@@ -1,8 +1,17 @@
-/* global describe */
+/* global describe, expect, it, shallow */
+import React from 'react';
 import * as common from '../../../test/unit/commonTests';
 import IconProject from './IconProject';
 
 describe('IconProject', () => {
     common.isConformant(IconProject, { tagName: 'svg' });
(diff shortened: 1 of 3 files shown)


src/components/Icon/IconProject.test.js
```diff
@@ -1,8 +1,17 @@
...
 import IconProject from './IconProject';
 
 describe('IconProject', () => {
     common.isConformant(IconProject, { tagName: 'svg' });
     common.addsClassName(IconProject, 'uir-icon');
+
+    it('sets title', () => {
+        const wrapper = shallow((
+            <IconProject title="Test title" />
+        ));
```
hkmarques: Don't really care, but just so it doesn't go unnoticed there's an extra `(` here.

jfrej: It's intentional. Otherwise the linter would complain about the lack of comma after the function argument. Using an additional set of `()` allows us to omit the comma after JSX.

PR #151, opened 2019-01-28
Add prop gradient to IconCompleted

FerreiraRaphael:

_None_

**New Features** 

Add property `gradient` to component `IconCompleted`, so that we can add a gradient color to icon when needed.

**Bug Fixes** 

_None_

### DIFF
--- a/src/components/Icon/IconComplete.js
+++ b/src/components/Icon/IconComplete.js
@@ -11,22 +11,30 @@ const propTypes = {
     className: PropTypes.string,
     style: StyleObjectPropType,
     title: PropTypes.string,
+    gradient: PropTypes.shape({
+        from: PropTypes.string.isRequired,
+        to: PropTypes.string.isRequired,
+    }),
 };
 
 const defaultProps = {
     className: null,
     style: null,
     title: 'Complete Icon',
+    gradient: null,
 };
 
 const IconComplete = ({
     className,
     style,
     title,
+    gradient,
     ...props
 }) => {
     lastInstanceId += 1;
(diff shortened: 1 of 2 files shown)


andyberry88:
Can we do this with CSS instead? If we change 1 icon we really should change them all to keep the APIs consistent. This may also require more thought about any implications.

jfrej:
Adding a very specific gradient like this (with fixed `gradientTransform` and gradient stops) might be too restricting for a fairly generic component library like ui-react unless we're sure that we always want to use the same gradient.

On the other hand adding props for every single gradient attribute to make it flexible would be overkill. We could take `griadientOpts` as a single prop, but it's still a bit messy.

Another option is to allow `gradient` prop which would take in a whole gradient react Node. That way everyone can use whatever gradients they like. But they would need to manually override styling and style the appropriate paths. This can be tricky because unfortunately our icons are a bit inconsistent.

As you can see there are several options and none of them are perfect. We'd need to investigate this further.

For now you can achieve the same effect outside of ui-react: it is OK to nest `<svg>` elements. So you can nest `<IconComplete>` within an `<svg>` with a gradient:

```
        <svg
            ...
        >
            <defs>
                {
                   <linearGradient id={gradientId} x1="-50%" x2="100%" gradientTransform="rotate(-40)">
                        <stop offset="0%" stopColor={gradient.from} />
                        <stop offset="60%" stopColor={gradient.to} />
                    </linearGradient>
                }
            </defs>
            <g>
                <IconComplete />
            </g>
        </svg>
```

Have a look at the `<IconProject>` component in the main repository.

I'm closing this because we need to investigate further.

PR #155, opened 2019-09-02
Add auto position prop to tooltip

Vazerthon:

- Changes the meaning of tooltip position names. `left` now means "to the left of the component" rather than "originating from the left edge"

**New Features** 

- Adds 'auto' value for tooltip position enum
- Positions tooltip as far from screen edge as possible

**Bug Fixes** 

_None_

**Performance Improvements** 

_None_

**Miscellaneous** 

Peek 2019-09-02 14-50

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -10,7 +10,7 @@ import TooltipBox from './TooltipBox';
 import { TooltipPosition } from './TooltipEnums';
 import './Tooltip.scss';
 
-export const propTypes = {
+const propTypes = {
     children: PropTypes.element.isRequired,
     className: PropTypes.string,
     position: ListPropType([
@@ -28,7 +28,7 @@ export const propTypes = {
     tooltip: ElementOrStringPropType.isRequired,
 };
 
-export const defaultProps = {
+const defaultProps = {
     className: null,
     position: TooltipPosition.TOP_CENTER,
     showTooltip: null,
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -1,3 +1,4 @@
+/* global window */
 import React, { Component } from 'react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
@@ -9,7 +10,7 @@ import TooltipBox from './TooltipBox';
 import { TooltipPosition } from './TooltipEnums';
 import './Tooltip.scss';
 
-const propTypes = {
+export const propTypes = {
     children: PropTypes.element.isRequired,
     className: PropTypes.string,
     position: ListPropType([
@@ -19,6 +20,7 @@ const propTypes = {
         TooltipPosition.BOTTOM_CENTER,
         TooltipPosition.BOTTOM_RIGHT,
         TooltipPosition.BOTTOM_LEFT,
+        TooltipPosition.AUTO,
     ]),
     showTooltip: PropTypes.bool,
     style: StyleObjectPropType,
@@ -26,7 +28,7 @@ const propTypes = {
     tooltip: ElementOrStringPropType.isRequired,
 };
 
-const defaultProps = {
+export const defaultProps = {
     className: null,
     position: TooltipPosition.TOP_CENTER,
     showTooltip: null,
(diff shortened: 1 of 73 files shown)


src/components/Tooltip/Tooltip.test.js
```diff
@@ -7,7 +7,18 @@ import TooltipBox from './TooltipBox';
 import Avatar from '../Avatar/Avatar';
 import AvatarCard from '../Avatar/AvatarCard';
 
-describe('Tooltip', () => {
+describe.only('Tooltip', () => {
```
jfrej:
```suggestion
describe('Tooltip', () => {
```

src/components/Tooltip/Tooltip.story.js
```diff
@@ -130,3 +131,38 @@ stories.add(
         </div>,
     ),
 );
+
+stories.add('Auto Position', () => (
+    <div>
+        <p>
+            Setting the `position` prop to `auto` or using `TooltipPosition.AUTO` allows the tooltip
```
jfrej: I think our enums are not exposed in the library exports. They are probably still accessible using direct path imports but it's better not to suggest it in the docs.

src/components/Tooltip/Tooltip.js
```diff
@@ -46,11 +66,15 @@ class Tooltip extends Component {
             : tooltip;
     }
 
-    handleFocus = () => this.setState({ showTooltip: true })
+    handleFocus = e =>
+        this.setState({ showTooltip: true, autoPosition: positionFromMouseEvent(e) });
```
jfrej: Does it still work with keyboard focus?

src/components/Tooltip/Tooltip.story.js
Vazerthon: I wasn't sure when I wrote this, thanks for confirming

src/components/Tooltip/Tooltip.js
Vazerthon: It does

src/components/Tooltip/Tooltip.js
```diff
@@ -46,35 +96,82 @@ class Tooltip extends Component {
             : tooltip;
     }
 
-    handleFocus = () => this.setState({ showTooltip: true })
+    getLeftPosition = position => ({
```
willcrisis: Instead of build the object  every time the function is called, isn't it better to define it in a `const` and then just reuse it on each call?

rhuankarlusmudano: I don't think that this would be that good, because a component can move from a place to another depending on the logic of the siblings components. Let's imagine the scenario where you have two components one with the tooltip and another that is inserted into the DOM depending on a state boolean flag. So for the first time you open the page this flag is **false** and only the component with the tooltip is shown on the page. The user interacts and with the app and this flag is changed to **true** in some way, and then, we can see both components on the screen. Given that this function calculates the tooltip position it will be updated and we will be able to see a tooltip on the updated right place which we will probably don't reach if we set this constant.

Vazerthon: I see what you're saying, it might be slightly more performant to have a static object but I think that'll be tricky to do because it's using state values.

I don't think there will be enough of a performance problem from this to make it worthwhile changing. What do you think?

williankrause: Not a big deal, TBH. You'll need to have a function which would receive the state values and build the object anyway. Maybe this overhead of building the object could be avoided with a `switch () { case... }` statement.

src/components/Tooltip/Tooltip.js
```diff
@@ -46,35 +96,82 @@ class Tooltip extends Component {
...
+            Math.floor((this.state.wrapperSize.height - this.state.tooltipSize.height) / 2),
+        [TooltipPosition.BOTTOM_RIGHT]: this.state.wrapperSize.height,
+        [TooltipPosition.BOTTOM_LEFT]: this.state.wrapperSize.height,
+        [TooltipPosition.BOTTOM_CENTER]: this.state.wrapperSize.height + this.props.tooltipOffset,
+        [TooltipPosition.TOP_RIGHT]: -(this.state.tooltipSize.height),
+        [TooltipPosition.TOP_LEFT]: -(this.state.tooltipSize.height),
+    })[position || TooltipPosition.TOP_LEFT];
+
+    handleFocus = e =>
+        this.setState({ showTooltip: true, autoPosition: positionFromFocusEvent(e) });
```
victorhtc: I know this is spread out in our codebase, but I'm not a fan of using arrow functions as class properties.
https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1
What's your view on this?

src/components/Tooltip/Tooltip.js
Vazerthon: Yeah, it was a `switch` for a while but I thought this was a little clearer and cleaner to read. I'm pretty easy either way though.

src/components/Tooltip/Tooltip.js
Vazerthon: Interesting. I hadn't seen that article before. I have no strong opinion on either way of declaring class properties tbh. I'm happy to change this to a function declaration.

I not too worried about any of the prototype/inheritance concerns from the article as we don't really do too much OO JS. I guess if there's a potential performance benefit we may as well take it though

PR #159, opened 2019-11-08
Textarea - Validation message under the field

rfontoura:
_None_

**New Features** 
Allow the use of a validation message that will be shown under the textarea field (instead of a tooltip for error)

**Bug Fixes** 
_None_

**Miscellaneous**
Adds an optional validation message under the field as an alternative to `tooltipError`
TextArea with validation message

1. Choose the type of validation message to `message` (the default is `tooltip`)
2. Provide a value for `validationMessage`
3. When the `isValid` is false the message will be shown under the textarea field

### DIFF
--- a/src/components/Fields/TextArea.js
+++ b/src/components/Fields/TextArea.js
@@ -16,6 +16,7 @@ const propTypes = {
     autoHideLabel: PropTypes.bool,
     className: PropTypes.string,
     componentRef: PropTypes.func,
+    errorMessageType: PropTypes.oneOf(['tooltip', 'message']),
     hasFixedHeight: PropTypes.bool,
     isDisabled: PropTypes.bool,
     isFullWidth: PropTypes.bool,
@@ -37,6 +38,7 @@ const propTypes = {
     tooltipError: ElementOrStringPropType,
     tooltipHint: ElementOrStringPropType,
     tooltipRequired: ElementOrStringPropType,
+    validationMessage: PropTypes.string,
     value: PropTypes.string,
 };
 
@@ -45,6 +47,7 @@ const defaultProps = {
     autoHideLabel: false,
     className: null,
     componentRef: null,
+    errorMessageType: 'tooltip',
     hasFixedHeight: false,
     isDisabled: false,
     isFullWidth: false,
@@ -66,6 +69,7 @@ const defaultProps = {
     tooltipError: null,
     tooltipHint: null,
     tooltipRequired: 'required',
+    validationMessage: null,
     value: '',
 };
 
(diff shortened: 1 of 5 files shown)


src/components/_variables.scss
```diff
@@ -10,4 +10,5 @@ $grey: #ccc !default;
 $charcoal: #2f2833 !default;
 $purple: #9a3cbb !default;
 $pink: #ff1966 !default;
+$pink2: #fc2268 !default;
```
Vazerthon: image

These colours are almost impossible to differentiate. I don't think we should be adding the new one

src/components/Fields/TextArea.js
```diff
@@ -262,38 +270,49 @@ class TextArea extends Component {
                 style={this.props.style}
                 {...proxyDataProps(this.props)}
             >
-                <div className="uir-text-area-inner">
-                    <div className="uir-text-area-label-wrapper">
-                        {label}
+                <div>
```
Vazerthon: Is this outer div needed?

src/components/Fields/TextArea.scss
```diff
@@ -107,9 +87,41 @@ $textFieldLabelColor: $charcoal !default;
...
+    }
+
     .uir-icon-required {
         width: 10px;
         height: 10px;
         margin-left: 5px;
     }
 }
+
+.uir-text-area-inner {
```
Vazerthon: Could this be a breaking change if any clients are overriding these styles?

src/components/Fields/TextArea.scss
```diff
@@ -107,9 +87,41 @@ $textFieldLabelColor: $charcoal !default;
         width: 100%;
     }
 
+    &-validation-message {
+        margin-top: 8px;
+        font-family: "DINOT", Arial, sans-serif;
+        font-size: 11px;
+        font-weight: normal;
+        color: $pink2;
+        letter-spacing: .92px;
```
Vazerthon: Does this letter spacing work? `.92px` seems like a suspicious value

src/components/Fields/TextArea.js
rfontoura: Yep. I hat do keep it to put the validation message outside the input field.

image

src/components/Fields/TextArea.scss
rfontoura: You mean they're using this same class name? I had to use it because it has to be applied to the innermost div to draw the pink underline under the field due to the addition of the external `div`. What are your suggestions for it?

src/components/Fields/TextArea.scss
rfontoura: Yes, it works. This property was defined per design. There's a subtle difference.

**With .92px letter-spacing**
image

**Without letter-spacing defined**
image

PR #165, opened 2019-11-13
Add on Open and Close event props to AvatarMenu Component

FerreiraRaphael:

_None_

**New Features** 

### #AvatarMenu new Props
- onMenuClose: () => void,
- onMenuOpen: () => void,

**Bug Fixes** 

_None_

**Performance Improvements** 

**Miscellaneous**

### DIFF
--- a/src/components/Avatar/AvatarMenu.js
+++ b/src/components/Avatar/AvatarMenu.js
@@ -18,6 +18,8 @@ const propTypes = {
         PropTypes.string,
     ]).isRequired,
     className: PropTypes.string,
+    onMenuClose: PropTypes.func,
+    onMenuOpen: PropTypes.func,
     open: PropTypes.bool,
     position: PropTypes.string,
     style: StyleObjectPropType,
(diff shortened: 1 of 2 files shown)


src/components/Avatar/AvatarMenu.js
```diff
@@ -18,6 +18,8 @@ const propTypes = {
         PropTypes.string,
     ]).isRequired,
     className: PropTypes.string,
+    onMenuClose: PropTypes.func,
+    onMenuOpen: PropTypes.func,
```
hkmarques: I'm itching to ask you to change the props declaration to the flow type notation, but I will not do that here.

src/components/Avatar/AvatarMenu.js
```diff
@@ -36,10 +40,19 @@ class AvatarMenu extends Component {
     }
 
     handleAvatarClick = () => {
-        this.setState({ open: !this.state.open });
+        const open = !this.state.open;
```
hkmarques: Just for clarity I think this would give a clearer impression of what it's doing if it was named `shouldOpen` or something similar. `open = !open` is a tad confusing.

FerreiraRaphael: You are right, I'm going to fix that

src/components/Avatar/AvatarMenu.js
hkmarques: Just realised this is ui-react! LOL

